PR #64, opened 2018-07-18
Js cancel save

joyastockwell:
So this just proves that, with the proper imports, we can instantiate classes from the back-end's temp-dev branch and implement their interfaces. I have a save button that calls the function in the dao interface that the save button ultimately will. It doesn't do anything with back-end yet. There's no cancel functionality yet.


sarabiaj:
We don't really need this in the app; classes to play around with functionality can be kept locally

PR #65, opened 2018-07-19
Back end dagger

Acliu1498:


src/main/kotlin/persistence/repo/Module.kt
```diff
@@ -1,24 +1,24 @@
...
-import persistence.repo.UserRepo
+import javax.inject.Singleton
 
-class TrDatabaseImpl: TrDatabase {
-    private val dataStore: KotlinEntityDataStore<Persistable>
-    private val userRepo: Dao<User>
-    private val languageRepo: Dao<Language>
-    init {
+@Module
+class Module {
```
sarabiaj: Module is too ambiguous of a class name for this

sarabiaj:
Also, dagger shouldn't be used to provide the datastore. Datastore is specific to the implementation of the persistence layer and is being injected into another thing specific to the implementation of the persistence layer, which means we're missunderstanding the purpose of a dependency injection framework.

src/main/kotlin/persistence/repo/Module.kt
mbr4477: Got it. That was changed to DaoModule

src/main/kotlin/persistence/repo/Component.kt
```diff
@@ -0,0 +1,12 @@
+package persistence.repo
+
+import dagger.Component
+import dagger.Provides
+import javax.inject.Singleton
+
+@Component(modules = [Module::class])
+@Singleton
+interface Component {
```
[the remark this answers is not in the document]
mbr4477: Changed to DaoComponent

src/main/kotlin/persistence/repo/Module.kt
sarabiaj: How about PersistenceModule?

PR #80, opened 2018-07-20
Pulling languages from door43 

jjhuang1:


src/main/kotlin/persistence/data/Api.kt
```diff
@@ -0,0 +1,24 @@
+package persistence.data
+
+import data.Language
+import retrofit2.Call
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+
+class Api {
```
sarabiaj: Api is too generic, we could have more than one API

src/main/kotlin/persistence/data/Api.kt
```diff
@@ -0,0 +1,24 @@
...
+    private val door43Api: Door43Api
+
+    init {
+        val retrofit = Retrofit.Builder()
+                .baseUrl("https://td.unfoldingword.org/")
```
sarabiaj: lets keep strings like this in a config file (resources file)

src/main/kotlin/persistence/data/Api.kt
```diff
@@ -0,0 +1,24 @@
...
+                .addConverterFactory(MoshiConverterFactory.create())
```
sarabiaj: Let's also use the RxJava2 adapter for retrofit, and return an observable instead of a Call object

src/main/kotlin/persistence/data/Door43Retrieval.kt
```diff
@@ -0,0 +1,46 @@
...
+        val execute = api.getLang("").execute()
+        return execute.body()
+    }
+
+    fun getNames() : Array<String>{
+        val execute = api.getLang("").execute()
+        var i = 0
+        var size = execute.body().size
+        var names = Array(size) { "n = $it" }
+        //var names : Array<String> = arrayOf<String>()
```
sarabiaj: please remove this commented out line

src/main/kotlin/persistence/data/Door43Retrieval.kt
```diff
@@ -0,0 +1,46 @@
...
+        }
+        return names
+    }
+
+    fun getSlugs() : Array<String>{
+        val execute = api.getLang("").execute()
+        var i = 0
+        var size = execute.body().size
+        var slugs = Array(size) { "n = $it" }
+        //var names : Array<String> = arrayOf<String>()
```
sarabiaj: same as above

src/main/kotlin/persistence/data/Door43Retrieval.kt
```diff
@@ -0,0 +1,46 @@
...
+            names[i] = langName.ln
+        }
+        return names
+    }
+
+    fun getSlugs() : Array<String>{
+        val execute = api.getLang("").execute()
+        var i = 0
+        var size = execute.body().size
+        var slugs = Array(size) { "n = $it" }
```
sarabiaj: if you absolutely need to use an Array, you can to .toArray from a list; this allows you to have a dynamically resizing list that you can just add to

src/main/kotlin/persistence/data/Door43Retrieval.kt
```diff
@@ -0,0 +1,46 @@
...
+        //var names : Array<String> = arrayOf<String>()
+        for(langName in execute.body()){
+            names[i] = langName.ln
+        }
+        return names
+    }
+
+    fun getSlugs() : Array<String>{
+        val execute = api.getLang("").execute()
+        var i = 0
```
sarabiaj: your C is showing lol.

src/main/kotlin/persistence/data/Lang.kt
```diff
@@ -0,0 +1,9 @@
+package persistence.data
+
+class Lang(
```
sarabiaj: this kind of thing should be a data class. I'd prefer it not be abbreviated. This class in general is also already being written by some other members of your team (perhaps try branching or merging with theirs)

src/main/kotlin/persistence/data/Lang.kt
```diff
@@ -0,0 +1,9 @@
...
+        val pk: Int, //id
+        val lc: String, //slug
+        val ln: String, //name
+        val gw: Boolean, //isGateway
+        val ang: String //anglicizedName
```
sarabiaj: these abbreviations are better just left unabbreviated

src/main/kotlin/persistence/data/Api.kt
jjhuang1: ok. changed to Door43Api and renamed the interface Door43Api to Door43ApiInterface

src/main/kotlin/persistence/data/Api.kt
jjhuang1: changed

src/main/kotlin/persistence/data/Api.kt
jjhuang1: added and changed

src/main/kotlin/persistence/data/Lang.kt
jjhuang1: I talked to my team members about adding the json name annotations to the language data class in the common repo (like @Json(name = "lc")val slug: String) so that I could use that class instead, but they wanted to keep the moshi specific annotations in the jvm repo, so I instead create another language class and then map it to the data class labels from the common repo. What do you think?

.gitignore
```diff
@@ -90,4 +90,12 @@ gradle-app.setting
 *.rar
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
-hs_err_pid*
\ No newline at end of file
+hs_err_pid*
+
+# ignore DS_STORE file
+.DS_STORE
```
sarabiaj: try *.DS_STORE

src/main/kotlin/persistence/data/Lang.kt
sarabiaj: this class doesn't have any annotations in it

But anyway, yes, I suppose they're right. 
First, let's move all the retrofit stuff into an "api" package rather than persistence.
Next, have a subpackage called "model" for this kind of thing. You'll then need a mapper class to map between the common data classes and yours.

src/main/kotlin/persistence/mapping/UserMapper.kt
```diff
@@ -0,0 +1,51 @@
...
+import data.Language
+import data.User
+import data.dao.Dao
+import io.requery.Persistable
+import io.requery.kotlin.eq
+import io.requery.sql.KotlinEntityDataStore
+import persistence.model.*
+import javax.inject.Inject
+
+class UserMapper @Inject constructor(private val dataStore: KotlinEntityDataStore<Persistable>, private val languageRepo: Dao<Language>, private val userPreferencesMapper: UserPreferencesMapper): Mapper<IUserEntity, User>{
```
sarabiaj: Line is too long, don't use @Inject here

src/main/kotlin/persistence/mapping/UserMapper.kt
```diff
@@ -0,0 +1,51 @@
...
+                .select(IUserLanguage::class)
+                .where((IUserLanguage::userEntityid eq type.id))
+                .get()
+                .toList()
+
+        val sourceLanguages = allUserLanguageJunctionTableRows.filter { it.source }
+                .map { languageRepo.getById(it.languageEntityid).blockingFirst() }
+        val targetLanguages = allUserLanguageJunctionTableRows.filter { !it.source }
+                .map { languageRepo.getById(it.languageEntityid).blockingFirst() }
+        val userPreferences = userPreferencesMapper.mapFromEntity(type.userPreferencesEntity)//dataStore.select(IUserPreferencesEntity::class).
```
sarabiaj: clear the comment

src/main/kotlin/app/ui/Application.kt
```diff
@@ -0,0 +1,8 @@
+package app.ui
```
sarabiaj: remove this file please

src/main/kotlin/persistence/JavaAppPreferences.kt
```diff
@@ -0,0 +1,18 @@
+/*package persistence
```
sarabiaj: remove this file please

mbr4477:
Outdated. Already in sprint1-merge

PR #81, opened 2018-07-20
Audio Storing

jjhuang1:
For takes and profile name recording wav files


src/main/kotlin/app/filesystem/AudioStore.kt
```diff
@@ -0,0 +1,55 @@
...
+    val locale = Locale.getDefault()
+    val labels= ResourceBundle.getBundle("Resources", locale)
+
+    val directoryProvider = DirectoryProvider(labels.getString("Application_Name"))
+
+    private val separator = FileSystems.getDefault().separator
+
+
+    //given the characteristics of a take, creates an audio file for that take in its project folder
+    fun createTakeFile(lang: String, anth: String, book: String, chap: String, chunk: String, take: String): File {
```
jsarabia: You really don't want to be passing in so many strings here. If anything changes at all about the spec or something, you have to change the parameters here, and then go hunt down EVERY place where you used it, and add or subtract the new parameters. You really would prefer to do something like take in a Take object, since that's going to be able to get all of this information for you.

That said, you actually would prefer that createFile be a function of the Take object itself, and provide to it the audio directory. Project should set up the intermediate directories, and take should just be responsible for the file itself.

src/main/kotlin/app/filesystem/AudioStore.kt
```diff
@@ -0,0 +1,55 @@
...
+        val path = lang + separator + anth + separator + book + separator + chap + separator + chunk + separator
+        val pathDir = directoryProvider.getUserDataDirectory(path, true)
+        val takeDirs = File(pathDir)
+        val fileSep = "_"
+        //make project directory to store take files in
+        takeDirs.mkdirs()
+        val takeFile = File(takeDirs,
+                "take"+fileSep+lang+fileSep+anth+fileSep+book+fileSep+chap+fileSep+chunk+fileSep+take+".wav")
```
jsarabia: this is a good candidate for string interpolation

src/main/kotlin/app/filesystem/AudioStore.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+
+    //given a file, checks that it is in the correct location and follows the naming convention for takes
+    fun checkTakeLocationAndName(audioTake: File): Boolean {
+        val absPath = audioTake.absolutePath
+        //audio take files follow convention "take_lang_anth_book_chap_chunk_takenum.wav"
+        return Pattern.matches(directoryProvider.getUserDataDirectory() +
+                separator+"[a-z]{2}"+separator+"[a-z]{2}"+separator+"[a-z]{3}"+separator+"\\d*"+separator+
+                "\\d*"+separator+"take_[a-z]{2}_[a-z]{2}_[a-z]{3}_\\d*_\\d*_\\d*.wav", absPath)
```
jsarabia: that user data directory method sure is nice; wouldn't it also be nice if you had a function like that to get you all those child directories?

you're also doing regex limited to very specific numbers of things. This is likely very easy to break since the method to generate the file is not as strict. things like chapter, chunk (or verse), and take in the previous file name convention had a letter at the end to make the numbers more human readable rather than just be like 03_03_01. This format also would not support verse ranges (current convention for chunk file names)

src/main/kotlin/app/filesystem/DirectoryProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+            success = true
+        } catch (e: Exception) {
+            // could not create the path or the directories
+            success = false
+        }
+        return success
+    }
+
+    // create a directory to store the user's application projects/documents
+    fun getUserDataDirectory(appendedPath: String = "", createIfNotExists: Boolean = true) : String {
```
jsarabia: why even give the option to not create if it doesn't exist?

src/main/kotlin/app/filesystem/AudioStore.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+
+    //creates a file in the internal storage of the given application to store the user's profile recording
+    fun createProfileRecordingFile(): File{
```
jsarabia: why limit this to audio? What if I want my profile configs here? user icon? etc
Why not let them just input their filename as a string?

src/main/kotlin/external_app.kt
```diff
@@ -0,0 +1,122 @@
...
+
+
+//gets the current status of Audacity on the user's system
+fun getCurrStatus(os: String, first: Boolean):Int{
+    val status_file: String
+
+    /*locates the appropriate shell executable file to execute in order to determine the
+      status of Audacity on the user's system*/
+    if (os.contains("WIN")) {
+        status_file = "./findstatus_WIN.sh"
```
jsarabia: I don't even know how you tested this as windows shouldn't natively run bash scripts?

src/main/kotlin/external_app.kt
```diff
@@ -0,0 +1,122 @@
...
+                process.start()
+            }
+            else { //Linux OS
+                val process = ProcessBuilder(program, take_recording_file.absolutePath)
+                process.start()
+            }
+            //calls helper function to get the current status of Audacity
+            var status = getCurrStatus(os, true)
+            while(status == 1){
+                status = getCurrStatus(os, false)
```
jsarabia: why not just use waitFor()? It should block the thread until the process returns. We could even have a dialog up that says "Please finish work in audacity before continuing" or something before you lock the UI.

src/main/kotlin/external_app.kt
```diff
@@ -0,0 +1,122 @@
...
+        //open Audacity (must be installed on the user's computer)
+        try{
+            //determines and executes the appropriate command to open Audacity
+            if (os.contains("WIN")){
+                //need to fix: Windows does not recognize start command for external apps
+                val process = ProcessBuilder("start", program, take_recording_file.absolutePath)
+                process.start()
+            }
+            else if (os.contains("MAC")){
+                val process = ProcessBuilder("open", "-a", program, take_recording_file.absolutePath)
```
jsarabia: Why not just use Runtime.exec?

src/main/kotlin/app/filesystem/AudioStore.kt
jjhuang1: So the Take object would have the language, anthology, book, chapter, chunk, and take as parameters?

src/main/kotlin/external_app.kt
jjhuang1: My bad, I didn't mean to include the audacity code in the pull request. I tried to test this on windows but couldn't figure out how to open Audacity from the command line without having to install some type of automation scripting tool... any ideas?

src/main/kotlin/external_app.kt
jjhuang1: I tried to use waitFor() on the command to open Audacity, but it kept on returning 0 right away and I assumed it was because it had already executed the command and audacity had already opened. I couldn't figure out how to get the exit code of audacity itself, so i just wrote getCurrStatus to go through the currently running processes and see if audacity was part of them or not to determine its status.

jsarabia: I don't know, I'd need to play around with it I guess. Your solution is definitely creative, I just don't know how fault tolerant it is (and maybe it is). One thing you probably would prefer to do for performance reasons is to not do a totally busy wait (it's going to be CONSTANTLY checking the status, and that status won't change frequently; you probably want to do a slight Thread.sleep for like, a second or a little less in between checks)

src/main/kotlin/external_app.kt
jsarabia: a .bat file would be my guess at an windows version of a bash script; but again, I'd want to do a little more research into this problem than to stick to a script that does a ps check.

joyastockwell:
Reformatting how we store things a little

jjhuang1:
@sarabiaj Image and audio storing functions have been tested, can you take another look at this?

src/main/kotlin/app/filesystem/Take.kt
```diff
@@ -0,0 +1,26 @@
...
+import java.util.*
+
+class Take (lang:String, anth: String, book: String, chap: String, chunk: String, take: String){
+
+    val locale = Locale.getDefault()
+    val labels= ResourceBundle.getBundle("Resources", locale)
+    val directoryProvider = DirectoryProvider(labels.getString("Application_Name"))
+
+    val fileSep = "_"
+    val filename = "take"+fileSep+lang+fileSep+anth+fileSep+book+fileSep+chap+fileSep+chunk+fileSep+take+".wav"
```
jsarabia: look into string interpolation

src/main/kotlin/app/filesystem/Take.kt
```diff
@@ -0,0 +1,26 @@
+package app.filesystem
+
+import java.io.File
+import java.nio.file.FileSystems
+import java.util.*
+
+class Take (lang:String, anth: String, book: String, chap: String, chunk: String, take: String){
```
jsarabia: prefer objects over strings here

src/main/kotlin/app/filesystem/ProfileConfig.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.io.File
+import filesystem.IDirectoryProvider
+
+// The User's Profile: the audio hash info in one file and an image in another file
+// Key for DirectoryProvider should be "Application_Name"
+class ProfileConfig(val directoryProvider: IDirectoryProvider) {
+    val locale = Locale.getDefault()
+
+    // Creates a user subdirectory and stores their hash info
+    fun storeAudio(userMine: User, audioFile: File): String {
```
jsarabia: natural language is important in naming conventions. UserMine is confusing since it's not something we'd say. CurrentUser? This also carries some connotations about state. Like, it matters that the user is currently logged in, or the method has some side effects because of the current logged in user. If that's not the case, just simply user is a better name here.

src/main/kotlin/app/filesystem/ProfileConfig.kt
```diff
@@ -0,0 +1,47 @@
...
+    // Returns path of new directory
+    fun createDirectory(directoryName: String): String {
+        //Todo: prevent illegal paths
+        val pathDir = directoryProvider.getAppDataDirectory(directoryName, true)
+        println("path: " + pathDir)
+        return pathDir
+    }
+
+    // Creates a file in the private internal storage of the given application
+    fun createFile(path: String, filename: String, fileGiven: File): String {
```
jsarabia: fileGiven is also not clear, similar to above

PR #82, opened 2018-07-23
Js time anim

joyastockwell:
This is now a widget that can be added to a widget folder instead of being a view all by itself.
The user can pass in color and time it takes to get filled in. I can change it so they can pass in size as well if we want different sizes.


mbr4477:
Closing for now

PR #83, opened 2018-07-23
Combo box widget

cabenedick:
Please enforce your wisdom on this code.


src/main/kotlin/app/ui/Application.kt
```diff
@@ -0,0 +1,5 @@
+package app.ui
+
+fun main(args: Array<String>) {
```
sarabiaj: let's not commit this file

src/main/kotlin/recources/UIColors.kt
```diff
@@ -0,0 +1,10 @@
+package recources
+
+object UIColors {
+
+    const val UI_PRIMARY = "#e56060" //ROMAN
```
sarabiaj: let's use an actual resources file (.properties) rather than a kotlin object

src/main/kotlin/widgets/LanguageChip.kt
```diff
@@ -0,0 +1,67 @@
...
+import tornadofx.*
+
+/**
+ * A language chip is a class used to store the data of the chip for
+ * easy direct access. This will allow access to the stackpane, and elements
+ * in the stackpane without searching through the children on the parent of the
+ * stackpane for what we want to access.
+ */
+
+class LanguageChip(val language: Language,
```
sarabiaj: if it's a widget, you should be creating a library component- as in, it should be generic, not specific to a use case within a specific system. As in, it shouldn't be dependent on a org.wycliffeassociates.data.Language object. Think about what a chip is, and only require the necessary information to create the view. Also, styling should ideally be separated from the view.

PR #108, opened 2018-07-26
Combo box widget

cabenedick:
We are ready for demoralization


src/main/kotlin/widgets/Chip.kt
```diff
@@ -0,0 +1,77 @@
...
+    val chip : StackPane
+
+    init {
+
+        label = label(labelText) {
+            textFill = textColor
+
+        }
+
+        deleteButton = button("X") {
```
sarabiaj: use an icon for this

src/main/kotlin/widgets/Chip.kt
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * A language chip is a class used to store the data of the chip for
+ * easy direct access. This will allow access to the stackpane, and elements
+ * in the stackpane without searching through the children on the parent of the
+ * stackpane for what we want to access.
+ */
+
+class Chip(val labelText : String,
+           fillColor : Color,
+           textColor : Color,
```
sarabiaj: Seems to me that colors would come through stylesheets, or type safe builders?

src/main/kotlin/widgets/Selector.kt
```diff
@@ -0,0 +1,178 @@
...
+import javafx.scene.paint.Color
+import tornadofx.*
+import java.util.*
+
+/**
+ * This class is used to make allow a user to select items from a filterable list
+ * and select which of those that are selected as the preferred, or default, item.
+ */
+
+class Selector(selectionData : List<DataSelectionInterface>,
```
sarabiaj: What selector?

src/main/kotlin/widgets/LanguageSelection.kt
```diff
@@ -0,0 +1,13 @@
+package widgets
+
+import data.model.Language
+
+class LanguageSelection(language : Language) : DataSelectionInterface {
```
sarabiaj: This is app specific; thus doesn't belong in the widgets package

cabenedick: Got it

src/main/kotlin/widgets/Selector.kt
cabenedick: Changed to DataListSelector, although I can change it to any more inspired name

cabenedick: Also changed the stylesheet name appropriately

src/main/kotlin/widgets/Chip.kt
k-horton: Done! ;)

src/main/kotlin/widgets/Chip.kt
k-horton: The problem with this is that the color is custom to the function of the comboBox (either target or source); because of that, it seems as though we either need to pass in the color or hard-code the color in depending on the function. Can you think of any other solution? We've thought long and hard about how to determine color and can't seem to find anything better... :(

src/main/kotlin/widgets/Selector.kt
```diff
@@ -0,0 +1,178 @@
...
+    private val viewModel = SelectorViewModel(updateSelections, preferredSelection)
+
+
+    private val chips = mutableListOf<Chip>()
+    private val selectionList = SelectionList(selectionData)
+    private val input = SimpleObjectProperty<String>()
+
+
+    private val colorNeutral = c(messages["UI_NEUTRAL"])
+    private val textFillNeutral = c(messages["UI_NEUTRALTEXT"])
```
sarabiaj: can we make this UI_NEUTRAL_TEXT?

src/main/kotlin/widgets/Selector.kt
```diff
@@ -0,0 +1,178 @@
...
+             */
+            editor.textProperty().addListener { _, _, newText -> this.setValue(newText) }
+
+            /**
+             * Check and see if a valid item is selected whenever the combobox dropdown is closed.
+             * If so, then save the item and make a chip for it.
+             */
+            addEventFilter(ComboBox.ON_HIDDEN) {
+                if (selectionList.observableList.contains(input.value)) {
+                    viewModel.addNewValue(selectionList.dataList[selectionList.observableList.indexOf(input.value)])
```
sarabiaj: line seems pretty long

src/main/kotlin/widgets/Selector.kt
cabenedick: Done

src/main/kotlin/widgets/Selector.kt
```diff
@@ -0,0 +1,178 @@
...
+
+                if (check.contains(true)) {
+                    chips.removeAt(check.indexOf(true))
+                } else {
+                    chips.add(0, Chip(
+                            selection.labelText,
+                            colorAccent,
+                            colorNeutral,
+                            viewModel::removeSelection,
+                            viewModel::newPreferredSelection))
```
sarabiaj:
```
chips.add(
    0, 
    Chip(
            selection.labelText,
            colorAccent,
            colorNeutral,
            viewModel::removeSelection,
            viewModel::newPreferredSelection
    )
)
```

src/main/kotlin/widgets/Selector.kt
```diff
@@ -0,0 +1,178 @@
...
+                children.addAll((chips.map { it.chip }))
+            })
+
+            /**
+             * Whenever a chip is clicked to be selected as default, change the color of the
+             * chips to represent that action.
+             */
+            compositeDisposable.add(preferredSelection.subscribe {
+                newSelected(it.labelText)
+            })
```
sarabiaj: styling

src/main/kotlin/widgets/Selector.kt
cabenedick: Moved selectionList.observableList.indexOf(input.value) into a variable (index) and passed that into viewModel.addNewValue(selectionList.dataList[ index ]) on the next line

src/main/kotlin/widgets/Selector.kt
cabenedick: Done

src/main/kotlin/widgets/Selector.kt
k-horton: It _was_ UI_NEUTRAL_TEXT before **someone** changed it @cabenedick

src/main/kotlin/widgets/DataSelectionInterface.kt
```diff
@@ -0,0 +1,12 @@
+package widgets
+
+/**
+ * This interface is used in Selector as the data from which is selected in the widget
+ */
+
+interface DataSelectionInterface {
```
sarabiaj: Poor name; how about something like "ComboBoxSelectionItem" or something

src/main/kotlin/widgets/Selector.kt
sarabiaj: additionally, the widget shouldn't be dependent on app resources

k-horton: How are you supposed to use the resource folder to store colors, then? D:

src/main/kotlin/widgets/DataSelectionInterface.kt
cabenedick: I'll go with that

src/main/kotlin/widgets/Chip.kt
cabenedick: They way the events happen, we actually don't need to set the colors inside the chip widget. Therefore, these parameters have been removed.

src/main/kotlin/widgets/Selector.kt
cabenedick: Passed those colors in as well as the accent color so that the widget is no longer dependent

cabenedick:
Rebased and applied suggested changes

src/main/resources/MyView.properties
```diff
@@ -0,0 +1,4 @@
+UI_PRIMARY = #e56060
```
sarabiaj: My_____ are names for tutorials only

k-horton: Refactored to "Application" resource bundle (would have called it "Language" but it's also storing colors right now)

cabenedick:
Closing pull request as comboBox_widget3 has been approved as a better alternative architecture version of this branch.

PR #110, opened 2018-07-27
Audio Playing

joyastockwell:


src/main/kotlin/app/ui/AudioPlayer.kt
```diff
@@ -0,0 +1,10 @@
+import javafx.scene.media.AudioClip
+
+fun playKotlin(URI: String) {
```
jsarabia: kotlin is not a thing that is "played"

src/main/kotlin/app/ui/AudioPlayer.kt
```diff
@@ -0,0 +1,10 @@
...
+    val myClip = AudioClip(URI)
+    myClip.play()
+    while(myClip.isPlaying) {
+        Thread.yield()
```
jsarabia: this is still a busy wait, and ultimately the player shouldn't be responsible for making this kind of thread management decision- AudioClip almost certainly spawns an internal thread to play on so what you're likely doing here is making decisions about the app's main thread.

src/main/kotlin/app/ui/AudioPlayer.kt
```diff
@@ -0,0 +1,10 @@
...
+    }
+    println("done playing")
```
jsarabia: We may want to introduce a logging class where things like this are done, but we probably don't want to be always writing things like this to standard out

joyastockwell:
Ok. I will get on this when I am done with a first pass at the file creation methods for the user's audio hash and clip.

joyastockwell:
Two questions:
1) I changed the function so it calls two helper functions: one to load from the URI and the other to play the loaded audio clip. Also, I want to put everything in a class so that I can mark that class as a singleton, but I don't know where I would put that class. Probably it shouldn't stay in the UI folder. Where's a good place for it?
2) Where should the threads get controlled from? Main? A thread controlling class? If the latter, should we have a separate branch for developing that?

PR #116, opened 2018-07-30
Combo box widget3

cabenedick:
This is a newer version of the architecture for comboBox_widget. I just wanted to see if you prefer this version over the other one. My main concern is the architecture itself; If you prefer the original version, then I'll scrap this one.

Main changes:

I made a combobox specific class (like how there is one for the chip).

I then moved the ComboBoxSelector logic into a view (Language selector). This way only widgets needed are the ComboboxSelector and the chip while the rest is app based and more flexible. 

Because there is now a app specific view that handles the logic, this also means that all chip color changes are handled in the view allowing us to use the resource properties to get the colors.

The view model and model were moved to the app since they are working with the Language Selector view. They now work with languages and combobox selection items, so the observables send language objects through them instead of the interface equivilant.


src/main/kotlin/widgets/ComboBoxSelector.kt
```diff
@@ -0,0 +1,51 @@
+package widgets
+
+import javafx.application.Platform
+import javafx.scene.control.ComboBox
+import tornadofx.*
+
+class ComboBoxSelector (
```
jsarabia: seems like what you're making here is more of a "SearchableComboBox" perhaps?

cBenedick: How about Filterable since all comboboxes are "searchable"

src/test/kotlin/app/ui/LanguageSelectorModel.kt
```diff
@@ -0,0 +1,37 @@
+package app.ui
+
+import data.model.Language
+import org.junit.Test
+
+class TestLanguageSelectorModel {
```
jsarabia: need to rename file

src/test/kotlin/app/ui/LanguageSelectorViewModel.kt
```diff
@@ -0,0 +1,68 @@
+package app.ui
+
+import data.model.Language
+import io.reactivex.subjects.PublishSubject
+import javafx.embed.swing.JFXPanel
+import org.junit.Test
+import widgets.Chip
+
+class TestLanguageSelectorViewModel {
```
jsarabia: need to rename file

src/test/kotlin/app/ui/LanguageSelectorModel.kt
cBenedick: Done

src/test/kotlin/app/ui/LanguageSelectorViewModel.kt
cBenedick: and done

src/main/kotlin/app/ui/LanguageSelectorViewModel.kt
```diff
@@ -0,0 +1,58 @@
...
+                setPreferredLanguage(model.preferredSelection)
+            }
+        }
+    }
+
+    private fun setPreferredLanguage(language: Language?) {
+        // we still want to notify the profile that there is no selected preferred language somehow?
+        if (language != model.preferredSelection) {
+            model.preferredSelection = language
+            if (language != null) {
```
jsarabia: you can use let syntax here

cBenedick: Changed the inner if to:
`language?.let {
    preferredLanguage.onNext(it)
}`

PR #137, opened 2018-08-02
Dev

Pinto18:
Changed script command for test stage in .travis.yml from 'gradlew test' to 'gradlew cleanTest test' to avoid the possibility of gradle skipping the tests due to lack of changes

Added modification to test task so that it prints out the name of each test and its status each time it is run.

More information for these changes can be found [here](https://stackoverflow.com/questions/3963708/gradle-how-to-display-test-results-in-the-console-in-real-time)


jsarabia (changes requested):
please merge only what you need

PR #140, opened 2018-08-03
clean up build.gradle

jsarabia:


Acliu1498 (approved):
its good, solved Nick's problem

PR #146, opened 2018-08-06
Languages Use Case

jjhuang1:
for inserting languages from door43 into the database


src/main/kotlin/usecases/GetLanguagesUseCase.kt
```diff
@@ -0,0 +1,29 @@
+package usecases
+
+import api.Door43Client
+import api.model.Door43Mapper
+import persistence.injection.DaggerDatabaseComponent
+
+class GetLanguagesUseCase {
```
jsarabia: This is more like a InitializeLanguageDatabaseUseCase

jsarabia (approved):

PR #147, opened 2018-08-06
Audio play

joyastockwell:
Plays audio from URI's and audio clips; separate functions for loading, playing, and stopping; function for loading and playing right after each other; tested on mac and windows


src/main/kotlin/device/audio/AudioPlayer.kt
```diff
@@ -0,0 +1,40 @@
+import javafx.scene.media.AudioClip
+import java.net.URI
+
+
+class AudioPlayer() {
+
+    var currAudioClip: AudioClip? = null
```
jsarabia: Why is this nullable?

joyastockwell: You can't initialize it as a class variable unless you assign it to something, so I set it to null.

src/main/kotlin/device/audio/AudioPlayer.kt
```diff
@@ -0,0 +1,40 @@
+import javafx.scene.media.AudioClip
+import java.net.URI
+
+
+class AudioPlayer() {
```
jsarabia: Why not have 3 constructors, one with a File, one with a String, and one with a URI? initialization with the audioplayer can be responsible for loading the audioclip.

Still, have a load() method (one for each of the three options) for the purpose of loading a new file.

Don't provide a "loadAndPlay()"

joyastockwell: Ok I'll get on that.

mbr4477: @sarabiaj If we are injecting the audio player, and it has a constructor that needs a File/String/URI, what will be the file we pass to the player in the Dagger module provides method?

src/main/kotlin/device/audio/AudioPlayer.kt
jsarabia: Why not start by initializing it?

joyastockwell: I changed it to be set on instantiation of AudioPlayer

mbr4477: @joyastockwell Did you push your changes?

src/main/kotlin/device/audio/AudioPlayer.kt
```diff
@@ -0,0 +1,21 @@
+import javafx.scene.media.AudioClip
+import java.net.URI
+
+
+class AudioPlayer(audioClip: AudioClip) {
```
jsarabia: AudioPlayer is meant to be a replacement for AudioClip in our app (because AudioClip is an implementation detail specific to the device you're on; Android uses AudioTrack). After all, if the developer already has an AudioClip object to use for constructing this, why wouldn't they just use AudioClip directly?

The API you're defining is to take in a:
- File
- URI (this is really more optional, but I'm fine with it being an option)
- String containing a path to a file

These are going to be defined in your interface, and implemented here.
You also want to use the constructor to take the opportunity for initializing your AudioClip. Since we have it there, there's no reason to make it a nullable. 

Nullable means the possibility of null pointer exceptions, the most common bug in all of software development. Kotlin forces you to do a null check to prevent this, but your elvis operators will mean that the code like play becomes a no-op if the player is null. Not playing when you call play though is a bug, but now you won't throw an exception, so the code doesn't believe something went wrong. We won't get a log of the stacktrace, and our user doesn't have working software.

We should be allergic to nullable types. There's a huge cost to allowing them, and it takes huge justification. If our current design/architecture won't allow us to avoid nullable types, we should first consider if there is another way of architecting the system so that they are no longer required. Only when we have exhausted all options should nullable be allowed.

mbr4477:
Closing to pull from an updated branch

PR #150, opened 2018-08-07
Audio recording

jjhuang1:


src/main/kotlin/device/audio/AudioRecorderImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+    private val audioByteObservable = PublishSubject.create<ByteArray>()
+    init {
+        line = AudioSystem.getTargetDataLine(FORMAT)
+    }
+
+    override fun start() {
+        line.open(FORMAT)
+        line.start()
+        Observable.fromCallable {
+            val byteArray = ByteArray(1024)
```
jsarabia: add buffer size to the companion object

src/main/kotlin/device/audio/AudioRecorderImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+            while (line.isOpen) {
+                totalRead += line.read(byteArray, 0, byteArray.size)
+                audioByteObservable.onNext(byteArray)
+            }
+        }
+                .subscribeOn(Schedulers.io())
+                .subscribe()
+    }
+
+    override fun stop() {
```
sarabiaj: if you're going to use an observable for recording, then you need to unsubscribe in stop, otherwise you leak the observable

src/main/kotlin/device/audio/AudioRecorderImpl.kt
```diff
@@ -0,0 +1,58 @@
...
+        line.start()
+        Observable.fromCallable {
+            val byteArray = ByteArray(BUFFER_SIZE)
+            var totalRead = 0
+            while (line.isOpen) {
+                totalRead += line.read(byteArray, 0, byteArray.size)
+                audioByteObservable.onNext(byteArray)
+            }
+        }
+                .subscribeOn(Schedulers.io())
```
jsarabia: line formatting

jsarabia (approved):

PR #151, opened 2018-08-07
Updated audio player

mbr4477:


mbr4477:
Travis will fail for now since we need the pr in common to go through with the interface

jsarabia (approved):

PR #155, opened 2018-08-07
Added LoginUseCase

mbr4477:


jsarabia (approved):

src/main/kotlin/usecases/LoginUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    fun playUserAudio(user: User) {
+        val audioFile = File(user.audioPath)
+        audioPlayer.load(audioFile)
+        audioPlayer.play()
+    }
+
+    fun getCurrentUser(): Observable<User> {
+        val userId = appPreferences.getCurrentUserId()
+        return if (userId != null)
```
jsarabia, line 39: use curly braces if your if else statement spans multiple lines. Also, you can use a .let here, or elvis

PR #166, opened 2018-08-22
Ns project home

natibekele:


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ViewModel/ProjectHomeViewModel.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.wycliffeassociates.otter.common.data.model.Project
+import org.wycliffeassociates.otter.common.domain.GetProjectsUseCase
+import tornadofx.*
+import org.wycliffeassociates.otter.common.data.model.Language
+import io.reactivex.Observable
+import java.util.*
+
+class ProjectHomeViewModel: ViewModel() {
+    // TODO(need repo)?
+    val getProjectsUseCase = GetProjectsUseCase()
```
natibekele: the GetProjectsUserCase has a Dao<Project> as a parameter. I'm drawing a blank as to what to do here... 
or do I end up injecting the usecase in the first place?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/View/ProjectHomeView.kt
```diff
@@ -0,0 +1,36 @@
+package org.wycliffeassociates.otter.jvm.app.ui.projecthome.View
```
jsarabia: lowercase packages

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/ProjectCard.kt
```diff
@@ -0,0 +1,18 @@
+package org.wycliffeassociates.otter.jvm.app.widgets
+
+import javafx.geometry.Pos
+import org.wycliffeassociates.otter.common.data.model.Project
+import tornadofx.*
+
+class ProjectCard(project : Project, openProject: ()-> Unit) : Fragment(){
```
jsarabia: Based on what I've read about creating JavaFX UI controls, you'd want to extend a layout here, not a Fragment, even though I suppose this should also work?

ultimately though, this is what should implement project card (it is the project card after all)

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ViewModel/ProjectHomeViewModel.kt
```diff
@@ -0,0 +1,15 @@
...
+    val projects : List<Project> = getProjectsUseCase.getProjects().toList().blockingGet().flatMap{it}
```
jsarabia: blocking get defeats the purpose of not blocking on the UI thread. 

You need to put this in an ObservableList, and bind that to the datagrid

When the list is populated because the observable has emitted, you update the ObservableList and datagrid will automatically update the view because of the observable binding

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/viewmodel/ProjectHomeViewModel.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.wycliffeassociates.otter.common.domain.GetProjectsUseCase
+import tornadofx.*
+import org.wycliffeassociates.otter.common.data.model.Language
+import io.reactivex.Observable
+import java.util.*
+
+class ProjectHomeViewModel: ViewModel() {
+    // TODO(need repo)?
+    var projectUseCase: GetProjectsUseCase? = null
+    val projects : Observable<List<Project>>? = projectUseCase?.getProjects()
```
jsarabia: How can we avoid a lateinit on projectusecase and prevent it from being nullable?

Can the viewmodel work if projectusecase is null? Nope, which then means the ability from any further things that now have to inherit nullable types are guaranteed to not work either. We need to avoid this even being possible.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/ProjectCard.kt
```diff
@@ -0,0 +1,30 @@
+package org.wycliffeassociates.otter.jvm.app.widgets
+
+import javafx.geometry.Pos
+import javafx.scene.layout.VBox
+import org.wycliffeassociates.otter.common.data.model.Project
+import org.wycliffeassociates.otter.common.ui.widgets.IProjectCard
+import tornadofx.*
+import tornadofx.Stylesheet.Companion.root
+
+class ProjectCard(project : Project, openProject: () -> Unit) : VBox(), IProjectCard{
```
jsarabia: Let's look at how click events are usually handled in UI controls for JavaFX- passing in a lambda doesn't seem to be consistent with how you use anything else.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/viewmodel/ProjectHomeViewModel.kt
```diff
@@ -0,0 +1,16 @@
...
+import tornadofx.*
+import org.wycliffeassociates.otter.common.data.model.Language
+import io.reactivex.Observable
+import org.wycliffeassociates.otter.jvm.app.ui.inject.Injector
+import java.util.*
+
+class ProjectHomeViewModel: ViewModel() {
+    // TODO(need repo)?
+    var projectUseCase = GetProjectsUseCase(Injector.projectDao)
+    val projects : Observable<List<Project>>? = projectUseCase?.getProjects()
```
jsarabia: doesn't need the elvis

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/ProjectCard.kt
```diff
@@ -0,0 +1,22 @@
+package org.wycliffeassociates.otter.jvm.app.widgets
+
+import javafx.geometry.Pos
+import javafx.scene.control.Button
+import org.wycliffeassociates.otter.common.data.model.Project
+import org.wycliffeassociates.otter.common.ui.widgets.IProjectCard
+import tornadofx.*
+import tornadofx.Stylesheet.Companion.root
+
+class ProjectCard(project: Project) : Button() {
```
jsarabia: a button is a control that you click, so the intent would be that the whole thing is a button. You can have image buttons, where part of the button has a picture, but the whole thing is just a button. That's not exactly what we want here though? Right now we have a couple labels, eventually it may be a label and an image, etc, but both then actually *contain* a button- implying the whole widget is not.

I'd extend VBox here, unless you know you want all of the APIs (or at least most of them) of the Button class.
(this also would mean you can pull the vbox out of the with root, since you become the vbox here)

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/view/ProjectHomeView.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.wycliffeassociates.otter.jvm.app.widgets.ProjectCard
+import tornadofx.*
+
+class ProjectHomeView : View() {
+    private val viewModel: ProjectHomeViewModel by inject()
+    val projects = viewModel.projects
+
+    override val root = borderpane {
+        center {
+            projects?.subscribeBy(
```
jsarabia: doesn't need an elvis

jsarabia (approved):

PR #191, opened 2018-08-31
Audio plugin, plugin parsing, and registrar

mbr4477:


src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/parser/ParsedAudioPluginDataMapper.kt
```diff
@@ -0,0 +1,32 @@
...
+        val osName = System.getProperty("os.name").toUpperCase()
+
+        // Get the executable for the system we are running on
+        val executable = if (osName.contains("WIN"))
+            parsedAudioPlugin.executable.windows
+        else if (osName.contains("MAC"))
+            parsedAudioPlugin.executable.macos
+        else parsedAudioPlugin.executable.linux
+
+        // Return the audio plugin or through an UnsupportedPlatformException
```
natibekele: *throw?

mbr4477: Yep, good catch

src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt
```diff
@@ -0,0 +1,26 @@
...
+                .fromCallable {
+                    // Build and start the process
+                    val process = ProcessBuilder(
+                            listOf(
+                                    pluginData.executable,
+                                    *(pluginData.args.toTypedArray()),
+                                    file.toString()
+                            )
+                    )
+                            .start()
```
jsarabia: styling

jsarabia: I guess the intent here was to tab off the ProcessBuilder, but it isn't clear when you look down vertically (especially since it lines up with the list of)

The specific guideline is ambiguous ("The first call in the chain usually should have a line break before it, but it's OK to omit it if the code makes more sense that way.") I think that having the .start on the end of the close parentheses is probably appropriate here

src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt
```diff
@@ -0,0 +1,26 @@
...
+                    process.waitFor()
+                }
+                .subscribeOn(Schedulers.io())
```
jsarabia: or do above the same as you did here

src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/parser/ParsedAudioPluginDataMapper.kt
```diff
@@ -0,0 +1,32 @@
...
+import java.io.File
+
+class ParsedAudioPluginDataMapper {
+    // Map from Jackson parser class to an AudioPlugin
+    // No need to map the other way
+    fun mapToAudioPluginData(parsedAudioPlugin: ParsedAudioPluginData, sourceFile: File): AudioPluginData {
+        val osName = System.getProperty("os.name").toUpperCase()
+
+        // Get the executable for the system we are running on
+        val executable = if (osName.contains("WIN"))
```
jsarabia: this should use a when statement

jsarabia (approved):

PR #208, opened 2018-09-10
Audio take card and audio player

mbr4477:
Needs common/mr_audio_player_updates to be merged into common/dev (updated IAudioPlayer interface)

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/takecard/Take.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/takecard/Take.kt
@@ -1,0 +1,11 @@
+package org.wycliffeassociates.otter.jvm.app.widgets.takecard
+
+import java.io.File
+import java.util.*
+
+data class Take(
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/takecard/Take.kt
```diff
@@ -0,0 +1,11 @@
+package org.wycliffeassociates.otter.jvm.app.widgets.takecard
+
+import java.io.File
+import java.util.*
+
+data class Take(
```
jsarabia, line 6: is there a reason this isn't in common, or using one from common?

mbr4477: We needed different things that was in the one in common. I put it here for now until we nail down exactly what properties should be in the class, at which point we can update the one in common.

mbr4477:
Moved to new, simpler custom UI control paradigm

PR #210, opened 2018-09-13
New take card

mbr4477:
## Usage

```kotlin
class SomeView : View() {
    override val root = vbox {
        val takes = Injector.takeDao.getTakesForContent(someContent)

        takes.forEach {
            takecard(it, Injector.audioPlayer)
        }
    }
}
```


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/Take.kt
```diff
@@ -0,0 +1,11 @@
+package org.wycliffeassociates.otter.jvm.app.widgets
+
+import java.io.File
+import java.util.*
+
+data class Take(
+        var number: Int,
+        var date: Date,
+        var file: File,
+        var played: Boolean = false
```
natibekele, line 10: could you add an ID to these Takes? Otherwise it's difficult to differentiate them when comparing takes.               
 
For the demo I did something where I added the Id to the data class and then added that Id as a property in the TakeModel. I then made it possible to get the take's id from the take card itself so that I could work with them inside the view.
 what do you think about that?

mbr4477: Could you use `number` as an unique identifier to differentiate? The `id` property we usually include is just for the database, not really meant to be used anywhere else. All the takes for verse/chunk should have unique numbers, though.

natibekele:
perhaps. I just ask because I know when working in tE I ran into takes that
had the same take number. That came from people starting projects on
different machines and then they end up importing them into the a shared
place where they clashed. Idk. I suppose we can leave it as is for now

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
```diff
@@ -0,0 +1,111 @@
...
+import javafx.scene.layout.Priority
+import javafx.scene.paint.Color
+import org.wycliffeassociates.otter.common.device.IAudioPlayer
+import org.wycliffeassociates.otter.jvm.app.UIColorsObject.Colors
+import tornadofx.*
+
+class TakeCard(take: Take, player: IAudioPlayer) : AnchorPane() {
+    private val badge = stackpane {
+        style {
+            backgroundColor += c(Colors["primary"])
```
natibekele: I believe @sarabiaj said that we don't want any of our widgets colors to be dependent on our project's resources?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
```diff
@@ -0,0 +1,111 @@
...
+            backgroundRadius += box(0.px, 10.px, 0.px, 10.px)
+            padding = box(8.px)
+        }
+        val icon = MaterialDesignIconView(MaterialDesignIcon.CREATION)
+        icon.size = "18px"
```
natibekele: you could also specify the size of the icon like this

`MaterialDesignIconView(MaterialDesignIcon.CREATION, "18px")`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
```diff
@@ -0,0 +1,111 @@
...
+            backgroundRadius += box(10.px)
+        }
+        val content = vbox {
+            style {
+                padding = box(10.px)
+            }
+            hbox {
+                vgrow = Priority.ALWAYS
+                style {
+                    spacing = 10.px
```
natibekele: spacing for the hbox can also be specified like so 

`hbox(Number) {} `

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
mbr4477: I'll double check, but I think the constructor for `MateriaDesignIconView` (vs `MaterialIconView`) doesn't support size

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
mbr4477: Good to know. Thanks!

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
natibekele: I'd be curious to know the answer to that actually

mbr4477: You had it right. Changed to `val icon = MaterialDesignIconView(MaterialDesignIcon.CREATION, "18px")`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
mbr4477: Should I just change them to default literal colors for now? Then we could apply our custom styles in the views

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/Take.kt
mbr4477: We will have an `id` field when the data model gets updated in common

I just realized though that even that id field will only be unique to the local user's database. Things like tE should end up creating their own unique ids when importing takes to make sure they can differentiate.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/TakeCard.kt
natibekele: Yes exactly!
@mbr4477 just move that out and it looks like we're good to go on the take cards. I want to get this merged and then pull new dev with take cards into view takes so that i can get that merged too

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/Take.kt
jsarabia: the take number won't be unique, it's a local number; id will be unique though because it would be re-assigned on import

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/SimpleAudioPlayer.kt
```diff
@@ -0,0 +1,111 @@
...
+        }
+    var pauseGraphic: Node? = null
+        set (value) {
+            field = value
+            if (isPlaying) {
+                playPauseButton.graphic = field
+            }
+        }
+
+    private var isPlaying = false
```
jsarabia: my suspicion is that this needs to be annotated @Volatile

jsarabia (approved):

PR #211, opened 2018-09-14
Updated Project Home with TFX MVVM

mbr4477:
- Removed use of unimplemented dao
- Added typesafe builder to project card
- Split into model, view, viewmodel
- Added filler data


mbr4477:
Close to pull directly into dev

PR #212, opened 2018-09-14
Refactor project home for TFX MVVM

mbr4477:
- Removed use of unimplemented dao
- Added typesafe builder to project card
- Split into model, view, viewmodel
- Added filler data


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/ProjectCard.kt
```diff
@@ -1,23 +1,31 @@
...
 import javafx.geometry.Pos
+import javafx.scene.control.Button
+import javafx.scene.layout.Pane
 import javafx.scene.layout.VBox
 import org.wycliffeassociates.otter.common.data.model.Project
 import tornadofx.*
 import tornadofx.Stylesheet.Companion.root
 
 class ProjectCard(project: Project) : VBox() {
+    val loadButton = Button("Load")
```
natibekele: the text for this button can't be hardcoded here. we need to localize it

mbr4477: Just leave it blank here and let the parent layout set the value?

natibekele: yes! I believe you should be able to do something like 

```
 projectcard { 
                             loadButton.apply {
                                text = messages["load"]
                               }
             } 
```
or whatever the resource key will be

natibekele (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/ProjectCard.kt
```diff
@@ -1,23 +1,31 @@
...
-    val loadProjectButton = button("LOAD")
     init {
         with(root) {
-
             vbox(20) {
                 alignment = Pos.CENTER
-                label(project.book.toString())
-                label(project.targetLanguage.toString())
-                add(loadProjectButton)
+                label(project.book.titleKey)
```
jsarabia, line 18: don't you need to do a resources lookup?

mbr4477: Correct. This was a placeholder since it will change once the new schema/data model classes are implemented. Should we go ahead and add the resources lookup now and just switch out where the key is coming from later?

jsarabia (approved):

PR #213, opened 2018-09-17
Ns view takes

natibekele:
new view that allows user to promote/ select takes for content


natibekele:
wait to rebase from new dev before merging

PR #231, opened 2018-10-03
Ns new wizard usecase

jsarabia:


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/model/ProjectCreationModel.kt
```diff
@@ -1,30 +1,77 @@
...
 
-    var resourceSelected : Collection by property()
-//    var resource by resourceProperty
-   // var resourceProperty = getProperty(ProjectCreationModel::resourceSelected)
+    val resources: Single<List<Collection>>
+        get() = creationUseCase.getSourceRepos()
+                .observeOnFx()
+    val all: Single<List<Collection>>
+        get() = creationUseCase.getSourceRepos()
+                .observeOnFx()
```
mbr4477, line 31: Do we need both of these if they are the same thing? Will they be different in the future?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/model/ProjectCreationModel.kt
```diff
@@ -1,30 +1,77 @@
...
+                }
+                .subscribe()
+    }
+
+    fun createProject() {
+        creationUseCase.newProject(Collection(selectedBook.sort,selectedBook.slug, "project",
+                selectedBook.titleKey ,selectedBook.resourceContainer)).observeOnFx().doOnSuccess {
+            creationUseCase.updateSource(it, selectedBook)
+
+        }.subscribe()
```
mbr4477: Might be able to reformat this a bit to be more readable with the chained calls

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/ProjectCreation.kt
```diff
@@ -3,19 +3,27 @@ package org.wycliffeassociates.otter.jvm.app.ui.projectcreation.view
...
-            MaterialIconView(MaterialIcon.COLLECTIONS_BOOKMARK, "16px"),
-            MaterialIconView(MaterialIcon.CROP_SQUARE, "16px"),MaterialIconView(MaterialIcon.BOOK, "16px"))
+            MaterialIconView(MaterialIcon.COLLECTIONS_BOOKMARK, "16px"), imageLoader(File(ClassLoader.getSystemResource("assets/Cross.svg").toURI())),
+          MaterialIconView(MaterialIcon.BOOK, "16px"))
     override  val canGoNext = currentPageComplete
+    override val canFinish= creationViewModel.allPagesComplete
+
     init {
         showStepsHeader = false
         showStepsHeader = false
```
mbr4477: Is this a duplicate line?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectBook.kt
```diff
@@ -1,27 +1,35 @@
...
 
 import javafx.geometry.Pos
+import org.wycliffeassociates.otter.jvm.app.UIColorsObject.Colors
 import org.wycliffeassociates.otter.jvm.app.ui.projectcreation.viewmodel.ProjectCreationViewModel
+import org.wycliffeassociates.otter.jvm.app.ui.projecthome.ProjectHomeView
 import tornadofx.*
 
-class SelectBook() : View() {
+
+class SelectBook : View() {
```
mbr4477, line 15: Seems like there are two types of select views: one for major sections like Bible/tW/tN and anthology, and the other for the specific project (which is a book in the case of the Bible). Is there anyway we could consolidate things into `SelectType` and `SelectProject` views? Just thinking out loud to see if that might work, but I don't know much about the Wizard component

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectBook.kt
```diff
@@ -1,27 +1,35 @@
...
+                        label(it.titleKey)
+                        label(it.labelKey)
+                        button(messages["select"])
                     }
+                }
             }
+//
+    override fun onSave() {
+        runAsync{
+            viewModel.createProject()
```
mbr4477: Might be able to put createProject call in a Completable instead of using `runAsync`, just to keep our approach to threading the same across the app

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectResource.kt
```diff
@@ -3,115 +3,79 @@ package org.wycliffeassociates.otter.jvm.app.ui.projectcreation.view.fragments
...
-                        text = messages["obs"]
-                        alignment = Pos.CENTER
-                        selectedProperty().onChange {
-                            if (it) {
-                                removeClass(ResourceStyles.unselectedCard)
-                                addClass(ResourceStyles.selectedCard)
-                            } else {
-                                removeClass(ResourceStyles.selectedCard)
-                                addClass(ResourceStyles.unselectedCard)
+                    viewModel.resourceList.onChange {
```
mbr4477: Does the toggle group need cleared first if the resourceList has existing items?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectResource.kt
```diff
@@ -3,115 +3,79 @@ package org.wycliffeassociates.otter.jvm.app.ui.projectcreation.view.fragments
...
+                return MaterialIconView(MaterialIcon.BOOK)
+            }
+            OBS.slug -> {
+                return imageLoader(File(ClassLoader.getSystemResource("assets/OBS.svg").toURI()))
+            }
+            TW.slug -> {
+                return imageLoader(File(ClassLoader.getSystemResource("assets/tW.svg").toURI()))
+            }
+        }
+        return MaterialIconView(MaterialIcon.COLLECTIONS_BOOKMARK)
```
mbr4477: Might be able to move this to the `else` clause of the when statement and do something like 
```
return when (...) {
    ULB.slug -> MaterialIconView(MaterialIcon.BOOK)
    OBS.slug -> ...
    ...
}
```

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/model/ProjectCreationModel.kt
```diff
@@ -1,30 +1,77 @@
...
     var sourceLanguageProperty: Language by property()
-
     var targetLanguageProperty: Language by property()
+    var selectedResource: Collection by property()
+    var selectedAnthology: Collection by property()
+    var selectedBook: Collection by property()
+    var anthologyList: ObservableList<Collection> by property(FXCollections.observableArrayList())
+    var bookList: ObservableList<Collection> by property(FXCollections.observableArrayList())
+
+    val languages: Single<List<Language>>
```
mbr4477: Would it be better to turn these into
```
val languages: ObservableList<Language> = FXCollections.observableList(mutableListOf())
...
init {
    creationUseCase
        .getAllLanguages()
        .observeOnFx()
        .subscribe { retrieved ->
            languages.addAll(retrieved)
        }
}
```
That might keep the viewmodel from having to do the assignment and subscribe to the observable, since I think we only get to retrieve the languages from the use case once at the beginning

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/viewmodel/ProjectCreationViewModel.kt
```diff
@@ -9,17 +16,58 @@ class ProjectCreationViewModel : ItemViewModel<ProjectCreationModel>(ProjectCrea
...
+    val bookList = FXCollections.observableArrayList<Collection>()
 
     val projectsProperty = bind(ProjectCreationModel::projectProperty)
+    val allPagesComplete = SimpleBooleanProperty(false)
+
+    init {
+
+        languagesListProperty.value.map {
+            languageList.setAll(it)
+        }.subscribe()
```
mbr4477: It might just be a style preference, but the subscribe/assign could be in the model. Then the view model can just expose the observable list from the model. 

Also, since observable lists are observable by default, you can skip wrapping them in properties. So if we want to expose the languages observable list from the model to the view, the view model could just have a member like `val languages = model.languages`.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeModel.kt
```diff
@@ -0,0 +1,36 @@
...
+
+        val allCollections : Single<List<Collection>>
+        get() =  projectUseCase.getAllRoot().observeOnFx()
+
+        val allProjects = FXCollections.observableArrayList<Collection>()
+
+    init {
+        allCollections.doOnSuccess{
+                allProjects.setAll(it)
+        }.subscribe()
```
mbr4477: Like before, we could change this to
```
val allProjects: ObservableList<Collection> = FXCollections.observableList()

init {
    projectUseCase
        .getAllRoot()
        .subscribe { retrieved ->
            allProjects.setAll(retrieved)
        }
}
```

mbr4477: or you could just call `getAllProjects` in the init block

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeView.kt
```diff
@@ -0,0 +1,65 @@
...
+class ProjectHomeView : View() {
+
+    val viewModel: ProjectHomeViewModel by inject()
+    override val root = borderpane {
+        top = hbox { alignment = Pos.CENTER_RIGHT
+            button("Refresh") {
+                prefWidth = 232.0
+                prefHeight = 40.0
+                action {
+                    viewModel.getAllProjects()
```
mbr4477, line 31: Would it be possible to move this to the `onDock` method to automatically refresh whenever the view appears?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeView.kt
```diff
@@ -0,0 +1,65 @@
...
+                }
+            }}
+        style {
+            setPrefSize(1200.0, 800.0)
+        }
+        center = datagrid(viewModel.allProjectsProperty.value) {
```
mbr4477: if we change allProjects to just be an observablelist instead of a property, we can do datagrid(viewModel.allProjects) instead. Wrapping in properties is good for strings, booleans, singular objects, etc., but observable lists should be fine for lists

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+
+     fun getAllProjects() = bind(ProjectHomeModel::getAllProjects)
+
+    fun createProject() {
+        find<ProjectCreationWizard> {
+            openModal()
+            onComplete {
+                getAllProjects()
+                //creationViewModel.createProject()
+                //workspace.dockInNewScope<ProjectHomeView>()
```
mbr4477: Probably remove these commented lines

src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioPlayer.kt
```diff
@@ -10,6 +10,7 @@ import org.wycliffeassociates.otter.common.device.IAudioPlayer
 import org.wycliffeassociates.otter.common.device.IAudioPlayerListener
```
mbr4477: For all these files that aren't part of your new stuff, but may have been changed during a rebase or something, try checking out the copies on dev: `git checkout dev -- src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioPlayer.kt` (or the whole folder, if you don't need any changes in it). Once you've checked out all the files you don't want to change, commit and push and they shouldn't show up in this pr anymore

src/main/kotlin/org/wycliffeassociates/otter/jvm/usecases/CreateProjectUseCase.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.wycliffeassociates.otter.common.data.model.Language
+import org.wycliffeassociates.otter.common.data.model.Resource
+import org.wycliffeassociates.otter.common.data.model.SourceCollection
+import org.wycliffeassociates.otter.jvm.persistence.repositories.CollectionRepository
+import org.wycliffeassociates.otter.jvm.persistence.repositories.LanguageRepository
+import org.wycliffeassociates.otter.jvm.persistence.repositories.ProjectRepository
+import org.wycliffeassociates.otter.jvm.persistence.repositories.SourceRepository
+
+
+class CreateProjectUseCase(val languageRepo: LanguageRepository,
```
mbr4477: It might be best to move this to the common repo. All the repo interfaces should be there, so this could go in the domain package in common.

src/main/kotlin/org/wycliffeassociates/otter/jvm/usecases/ProjectUseCase.kt
```diff
@@ -0,0 +1,20 @@
+package org.wycliffeassociates.otter.jvm.usecases
+
+import io.reactivex.Single
+import org.wycliffeassociates.otter.common.data.model.Collection
+import org.wycliffeassociates.otter.jvm.persistence.repositories.ProjectRepository
+
+class ProjectUseCase(val projectRepo : ProjectRepository) {
```
mbr4477: Also can likely move to domain package of common

src/test/kotlin/org/wycliffeassociates/otter/jvm/persistence/TestDataStore.kt
```diff
@@ -3,6 +3,7 @@ package org.wycliffeassociates.otter.jvm.persistence
 import org.wycliffeassociates.otter.common.data.model.*
```
mbr4477: I'd just recheckout the test folder from dev, if you didn't add anymore tests. All lot of stuff got removed, so that should dramatically reduce the files that show up in this pr. `git checkout dev -- src/test`. Just make sure you've fetched the latest dev.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/ProjectCreation.kt
natibekele: if you're referring to the can coGoNext and can Finish, no, the canGoNext allows you to progress in the wizard. the canFinish activates the finish button once all the pages have been completed by the user.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/model/ProjectCreationModel.kt
natibekele: no, that was something that i was using before and missed on clean up. thanks

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/ProjectCreation.kt
natibekele: oh i see what you're talking about now, sorry. yeah it is a duplicate

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectResource.kt
natibekele: no. i don't think so. the call to get the resources only happens once initially when the wizard is initialized. and since the response is a single it shouldn't push out any more changes after the initial one i believe

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+import tornadofx.*
+
+class ProjectHomeViewModel : ItemViewModel<ProjectHomeModel>(ProjectHomeModel()) {
+
+    val allProjectsProperty = bind(ProjectHomeModel::allProjects)
+
+     fun getAllProjects() = bind(ProjectHomeModel::getAllProjects)
+
+    fun createProject() {
+        find<ProjectCreationWizard> {
```
mbr4477: Maybe this logic should go in the model?

mbr4477: I ended up have the viewmodel pass the Workspace back to the model so the model could dock/launch the next view. Not sure if that is necessary here

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/model/ProjectCreationModel.kt
natibekele: so much better. thanks

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeView.kt
natibekele: yes. it will go in the onDock. reason i have that button there for right now is i was having problems getting the projects to refresh on the homepage once the wizard closes. the wizard was closing and docking the new view before the database had been updated with the new project, so i couldn't the getAllProjects wasn't getting all the projects at the time. I figured i'd sandbox myself on that, get other stuff done and then come back to it once everything else was filled out

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeView.kt
natibekele: yes, but the bind delegate from tornadofx uses properties by default does it not?

mbr4477: True, but with an observable list you won't need to use bind actually. You can just do `val allProjects = model.allProjects` (or similar) in the view model. We can already bind directly to an observable list. We were making custom properties for other fields because they were not observable by default.

natibekele: yes you're right. Problem i'm having right now is i'm not actually initializing a model in the view model. i'm using some ItemViewModel from tornadofx, but i suppose i can go ahead and change it. might have to do that for the wizard too but i'll go give it a shot real quick

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectBook.kt
natibekele: i can just remove that. i don't believe i need that async

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeView.kt
mbr4477: Could you do something like `val allProjects = item.allProjects`? I think item is property in ItemViewModel

mbr4477 (approved):

build.gradle
```diff
@@ -34,7 +34,7 @@ plugins {
     id 'nu.studer.jooq' version '3.0.1'
 }
 
-group 'org.wycliffeassociates.translationrecorder'
```
sarabiaj: why was this added?

src/main/Resources/assets/Old Testament (2).svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><defs><style>.cls-1{fill:none;opacity:0;}</style></defs><title>Old Testament</title><g id="Layer_2" data-name="Layer 2"><g id="Layer_1-2" data-name="Layer 1"><rect class="cls-1" width="24" height="24"/><path d="M22,20H14.94V14.25h3.83a.45.45,0,0,0,.46-.44.45.45,0,0,0-.46-.44H14.94v-.89h3.83a.44.44,0,1,0,0-.88H14.94v-.93h3.83a.45.45,0,1,0,0-.89H14.94A6.5,6.5,0,0,0,12.71,5,6.1,6.1,0,0,1,16,4a5.9,5.9,0,0,1,6,5.78Zm-8,0H2C2,16.54,2,13.09,2,9.63A6.08,6.08,0,0,1,8.11,4,6.11,6.11,0,0,1,14,9.78ZM8,5.78a4.22,4.22,0,0,0-4.1,4v8.44h8.3c0-2.85.06-5.7,0-8.55A4.17,4.17,0,0,0,8,5.78ZM7.57,16H5.72a.45.45,0,1,1,0-.89H7.57a.45.45,0,1,1,0,.89Zm2.77-1.7H5.72a.45.45,0,0,1-.46-.44.45.45,0,0,1,.46-.44h4.62a.45.45,0,0,1,.46.44A.45.45,0,0,1,10.34,14.25Zm0-1.77H5.72a.44.44,0,1,1,0-.88h4.62a.44.44,0,1,1,0,.88Zm0-1.81H5.72a.45.45,0,1,1,0-.89h4.62a.45.45,0,1,1,0,.89Z"/></g></g></svg>
```
jsarabia: any reason for the (2)? can we make the spaces "-" or "_"?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectBook.kt
jsarabia: Do you mean like a generic Select<T> view?

src/main/Resources/assets/Old Testament (2).svg
natibekele: no that's just how it was locally after download. (2) is there bc i downloaded it multiple times

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projecthome/ProjectHomeModel.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.wycliffeassociates.otter.common.data.model.Collection
+import org.wycliffeassociates.otter.jvm.app.ui.inject.Injector
+import org.wycliffeassociates.otter.common.domain.usecases.ProjectUseCase
+import org.wycliffeassociates.otter.jvm.app.ui.projectcreation.view.ProjectCreationWizard
+import tornadofx.*
+
+class ProjectHomeModel {
+    val projectUseCase = ProjectUseCase(Injector.projectRepo)
+
+        val allProjects = FXCollections.observableArrayList<Collection>()
```
jsarabia: unneeded indent

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/fragments/SelectResource.kt
```diff
@@ -3,147 +3,71 @@ package org.wycliffeassociates.otter.jvm.app.ui.projectcreation.view.fragments
...
-//                wizardcard {
-//                    text = "Bible"
-//                    selected = true
-//                    // image = imageLoader(File("/Users/NathanShanko/Downloads/OBS.svg"))
-//                }
-                togglegroup {
-
-                    addEventHandler(ActionEvent.ACTION) {
-                    }
+        override val complete = viewModel.resourceSelected
```
jsarabia: indentation

PR #235, opened 2018-10-09
Refactor filterable combobox from 8woc

mbr4477:
**All the deletions are from removing outdated 8woc profile creation code that broke when the combobox was refactored, and other user creation views that are outdated.** If we return to the profile screens in the future, we can checkout these files from a previous dev commit.

## Usage
```kotlin
filterablecombobox<Language> {
    // Set up the string converter
    converter = object: StringConverter<FilterableItem<Language>>() {
        override fun toString(filterable: FilterableItem<Language>?): String
                = filterable?.let { "${it.item.name} (${it.item.slug})" } ?: ""

        override fun fromString(string: String): FilterableItem<Language>?
                = items.filter { string.contains(it.item.slug) }.firstOrNull()
    }

    Injector
            .languageRepo
            .getAll()
            .subscribe { retrieved ->
                items = FXCollections.observableList(
                        retrieved.map { 
                            FilterableItem(it, listOf(it.name, it.anglicizedName, it.slug)) 
                        }
                )
            }

    selectionModel.selectedItemProperty().onChange {
        println(it?.item)
    }
}
```


natibekele (dismissed):

mbr4477:
Updated usage
```kotlin
filterablecombobox<Language> {
    // Set up the string converter
    converter = object: StringConverter<Language>() {
        override fun toString(language: Language?): String
                = language?.let { "${it.name} (${it.slug})" } ?: ""
        override fun fromString(string: String): Language?
                = items.filter { string.contains(it.slug) }.firstOrNull()
    }

    items.onChange { updatedLanguages ->
        filterItems.setAll(
            updatedLanguages.list.map { 
                // Create the filterable item with the filter parameters
                FilterableItem(it, listOf(it.name, it.anglicizedName, it.slug))
            }
        }
    }

    selectionModel.selectedItemProperty().onChange {
        println(it?.item)
    }
}
```

Keeping converter since the exact conversion from object to string is dependent on the generic type and won't be the same for all filterable boxes (trying to treat this as a generic box, not just for languages).

mbr4477:
Added `filterConverter` function that maps an item to the filter key words. No more need to mess with FilterableItem in the usage.
```kotlin
filterablecombobox<Language> {
    // Set up the string converter
    converter = object: StringConverter<Language>() {
        override fun toString(language: Language?): String
                = language?.let { "${it.name} (${it.slug})" } ?: ""
        override fun fromString(string: String): Language?
                = items.filter { string.contains(it.slug) }.firstOrNull()
    }

    filterConverter = { language ->
        listOf(language.name, language.slug, language.anglicizedName)
    }

    selectionModel.selectedItemProperty().onChange {
        println(it?.item)
    }
}
```

PR #261, opened 2018-10-18
Dao updates, import RC zip files

mbr4477:
- Uses new DaoProcs class to handle db transactions involving multiple table daos
- Allows import option from folder or zip file

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/CollectionRepository.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/CollectionRepository.kt
@@ -7,11 +7,11 @@ import io.reactivex.schedulers.Schedulers
 import org.wycliffeassociates.otter.common.data.model.Collection
 import org.wycliffeassociates.otter.common.data.model.RelatedCollectionContent
 import org.wycliffeassociates.otter.common.data.model.ResourceMetadata
-import org.wycliffeassociates.otter.common.domain.usfm.UsfmDocument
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/CollectionRepository.kt
```diff
@@ -103,6 +106,29 @@ class CollectionRepository(
                 .subscribeOn(Schedulers.io())
     }
 
+    override fun duplicateCollectionAndContent(obj: Collection, newMetadata: ResourceMetadata): Single<Collection> {
```
jsarabia, line 109: The repository is your public API, which means, it should perform high level abstract concepts that the app developer would be trying to do from the domain layer. 

The complex naming of this method is a sign that this functionality isn't exactly the high level intent that the app developer has. Why would they want to duplicate collections and content? Do they want to create a derivative project, in which one of the intermediate steps is to copy the source project's structure? Should that intermediate step not be private?

Method signatures are also much more important in public APIs meant to be used by other teams than they are in internal private method blocks (I get that we're one team here, and we often work in multiple layers, but as the repo developer you are on the database team, providing a service to the app team). There is no API manual, the method name is unclear, so names of parameters are the next step at deciphering the intention of the method. Object is as useless a name as "thing" here because basically everything in kotlin is an object. Is this the source project? Collection is also ambiguous since there's different types- "book," "chapter," "help," "manual" are some potential types here and may have implications on what kind of actions you're looking to do.

mbr4477:
Architecture change

PR #285, opened 2018-10-24
Update audio plugin directory and take directory path

mbr4477:
**Depends on common/mr-plugin-audio-dirs**


.dsp_cache
```diff
@@ -0,0 +1,3 @@
+info=<Dirac library cache file - do not edit>
```
jsarabia: do we need to add .dsp_cache to get ignore?

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/DirectoryProvider.kt
```diff
@@ -27,37 +27,45 @@ class DirectoryProvider(private val appName: String) : IDirectoryProvider {
...
+    override fun getProjectAudioDirectory(
+            book: Collection,
+            chapterNumber: Int
+    ): File {
+        // <user data directory>/{language slug}/{rc slug}/{book slug}/{%02d, chapter number}/
+        val appendedPath = listOf(
+                book.resourceContainer?.language?.slug ?: "no_language",
+                book.resourceContainer?.identifier ?: "no_rc",
+                book.slug,
+                "%02d".format(chapterNumber)
```
jsarabia: this is also affected by the %03 thing; perhaps chapter number should just be provided as a preformatted string?

PR #289, opened 2018-10-25
Remove takes deleted from filesystem

mbr4477:
**Depends on common/mr-remove-deleted-takes**

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/MarkerMapper.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/MarkerMapper.kt
@@ -13,10 +13,10 @@ class MarkerMapper {
         )
     }
 
-    fun mapToEntity(type: Marker): MarkerEntity {
+    fun mapToEntity(type: Marker, takeFk: Int? = null): MarkerEntity {
         return MarkerEntity(
                 type.id,
-                null,
+                takeFk,
                 type.number,
                 type.position,
                 type.label
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/TakeMapper.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/TakeMapper.kt
@@ -19,10 +19,10 @@ class TakeMapper {
         )
     }
 
-    fun mapToEntity(obj: Take): TakeEntity {
+    fun mapToEntity(obj: Take, contentFk: Int = -1): TakeEntity {
         return TakeEntity(
                 obj.id,
-                -1, // Not a valid value. Must be overridden before DB insert
+                contentFk,
                 obj.filename,
                 obj.path.toURI().path,
                 obj.number,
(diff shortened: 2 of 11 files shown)


natibekele (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/TakeMapper.kt
```diff
@@ -22,7 +22,7 @@ class TakeMapper {
     fun mapToEntity(obj: Take): TakeEntity {
         return TakeEntity(
                 obj.id,
-                null,
+                -1, // Not a valid value. Must be overridden before DB insert
```
jsarabia: Is there a time when we're mapping and don't have the chunk id?

mbr4477: Mapping from Take to TakeEntity, we do not have the id of the chunk in the context of the mapper. Although the Take does always belong to another chunk, that information is not stored in the Take object.

An alternative solution would be to pass the owning chunk (or even just the id of that chunk) into the mapper function as well and initialize the TakeEntity with that chunk's id. Currently, the chunk id is set in the take repository after the mapper has been called.

jsarabia: I think I prefer passing in the chunk to the mapper

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/MarkerMapper.kt
```diff
@@ -13,10 +13,10 @@ class MarkerMapper {
         )
     }
 
-    fun mapToEntity(type: Marker): MarkerEntity {
+    fun mapToEntity(type: Marker, takeFk: Int? = null): MarkerEntity {
```
jsarabia, line 16: I'd prefer to have the actual take object passed in, and then use take.id

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/TakeMapper.kt
```diff
@@ -19,10 +19,10 @@ class TakeMapper {
         )
     }
 
-    fun mapToEntity(obj: Take): TakeEntity {
+    fun mapToEntity(obj: Take, contentFk: Int = -1): TakeEntity {
```
jsarabia, line 22: same as above, require passing in the content

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/mapping/MarkerMapper.kt
mbr4477: There are instances where we have the takeFk, but not the take object itself to pass in. In those instances I think we have the TakeEntity, so we could pass in the TakeEntity instead of just the id, if that would be better.

jsarabia: can you point me to such instances?

mbr4477: Update method of TakeRepository, CollectionRepository, and ChunkRepository. We fetch the entity so we don't overwrite the existing relationship data, but we never have the full object

jsarabia: I suppose these are just more internal private methods anyway

jsarabia (approved):

PR #295, opened 2018-10-31
Derive project in DAO transaction

mbr4477:
**Depends on common/mr-derive-project**

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/inject/Injector.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/inject/Injector.kt
@@ -23,9 +23,6 @@ object Injector {
     val languageRepo = LanguageRepository(database, LanguageMapper())
     val collectionRepo = CollectionRepository(
             database,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/inject/Injector.kt
```diff
@@ -21,7 +21,13 @@ object Injector {
     val directoryProvider = persistenceComponent.injectDirectoryProvider()
     val resourceContainerDirectory = directoryProvider.resourceContainerDirectory
     val languageRepo = LanguageRepository(database, LanguageMapper())
-    val collectionRepo = CollectionRepository(database, CollectionMapper(), ResourceMetadataMapper(), LanguageMapper())
+    val collectionRepo = CollectionRepository(
+            database,
+            CollectionMapper(),
```
jsarabia: Why don't we move directoryProvider up in the arguments list, and allow the default arguments to construct the Mappers

jsarabia (approved):

PR #296, opened 2018-10-31
Ns block project creation

natibekele:


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/model/ProjectCreationModel.kt
```diff
@@ -62,7 +65,7 @@ class ProjectCreationModel {
     fun doOnUserSelection(selectedCollection: Collection, workspace: Workspace) {
         if (selectedCollection.labelKey == "book") {
             createProject(selectedCollection)
-            workspace.dock<ProjectHomeView>()
+            showOverlayProperty.set(true)
```
mbr4477: I think it might be better to insert this between lines 102 & 103 as part of the createProject function just inside in the null check. In the off chance something were to happen and the language was null, the overlay would get stuck on the screen.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectcreation/view/ProjectCreation.kt
```diff
@@ -58,6 +60,13 @@ class ProjectCreationWizard : Wizard() {
         add(SelectLanguage::class)
         add(SelectCollection::class)
 
+        creationViewModel.creationCompletedProperty.onChange {
+            if (it) {
+                runLater {
+                    close()
+                }
+            }
+        }
```
mbr4477, line 72: Does this still work if the user tries the wizard again? I didn't see where the property is reset to false

natibekele: i'm not certain. i'll check

mbr4477 (approved):

PR #297, opened 2018-11-05
Jbs fast rc import

jsarabia:


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/CollectionRepository.kt
```diff
@@ -238,4 +243,51 @@ class CollectionRepository(
...
+            is Tree -> {
+                importCollection(parentId, metadataId, node, dsl)
+            }
+            is TreeNode -> {
+                importChunk(parentId, node, dsl)
+            }
+        }
+    }
+
+    private fun importCollection(parentId: Int, metadataId: Int, node: Tree, dsl: DSLContext){
```
mbr4477: If you change parentId to nullable (which is fine since parentFk can be null), you can just call this function in `importResourceContainer` to import the root collection, passing in a null parentId

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/CollectionRepository.kt
```diff
@@ -238,4 +243,51 @@ class CollectionRepository(
...
+        val collection = node.value as Collection
```
mbr4477: If we want to make this more type safe, you could wrap this function code in a smart cast. Could do the same in import chunk.
```kotlin
val collection = node.value
if (collection is Collection) {
    // ...
}
```

mbr4477 (approved):

PR #301, opened 2018-11-07
Mr faster derive project

mbr4477:


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/CollectionRepository.kt
```diff
@@ -238,4 +364,53 @@ class CollectionRepository(
 
         return collectionMapper.mapFromEntity(entity, metadata)
     }
+    override fun importResourceContainer(rc: ResourceContainer, rcTree: Tree, languageSlug: String): Completable {
+        return Completable.fromAction {
+            val totalTime = measureTimeMillis {
```
jsarabia: let's remove this now that we don't need the profiling

jsarabia (approved):

PR #315, opened 2018-11-13
Change resource access from file to stream. Modify gradle for jar build

danparisd:
Includes Matthew's changes for file access to stream. I modified build.gradle to build an executable jar.


build.gradle
```diff
@@ -79,7 +82,7 @@ dependencies {
     compile "no.tornado:tornadofx:$tornado_version"
 
     //SVG Loader
-    implementation 'com.github.afester.FranzXaver:Examples:0.1'
+    compile 'com.github.afester.FranzXaver:Examples:0.1'
```
mbr4477: @danparisd Were the dependencies not working with `implementation`? Not 100% sure when, but I think `compile` is being deprecated in favor of `implementation`.

build.gradle
```diff
@@ -118,9 +121,34 @@ dependencies {
...
+    }
+    from (configurations.compile.collect { zipTree(it) }) {
+        exclude "META-INF/MANIFEST.MF"
+        exclude "META-INF/*.SF"
+        exclude "META-INF/*.DSA"
+        exclude "META-INF/*.RSA"
+    }
+}
+
+jfx {
```
mbr4477: What is this task for? I understand the `jar` one, but not quite sure what the `jfx` one does

build.gradle
danparisd: @mbr4477 Correct, the dependencies were not working with implementation because implementation does not expose the methods in the dependency, while compile does. 

https://stackoverflow.com/questions/44493378/whats-the-difference-between-implementation-and-compile-in-gradle

Compile is deprecated, so it would seem that compile might ought to be replaced with API, but compile was already being used in the build.gradle so I just went with compile because I knew it would work and it was simpler.

https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

https://www.youtube.com/watch?v=7ll-rkLCtyk&feature=youtu.be&t=1647

I don't know what it would take to allow implementation to be used in all the places compile is used, or if that's even viable. I'm open to replacing compile with api, but I'm not clear on how that benefits us, other than using the more modern configuration. Let me know your thoughts.

build.gradle
danparisd: @mbr4477 I actually just copied that from one of @sarabiaj gradle builds he was using while mine was breaking. I'm not sure if it's necessary because I don't think we're using any of the jfx tasks. I will experiment with removing it. Joe is there a reason I should keep it that you know of?

mbr4477: Joe mentioned that this might be important when we encounter systems that don't have JavaFX bundled with the JDK (e.g., ARM)

build.gradle
mbr4477: @danparisd Is `implementation` an issue because we need the class paths exposed for the `jar` task?

jsarabia: let's take the opportunity to get this done the right way, even if we had some compiles before.

I'd rather not move backwards, knowing we already need to fix it (deprecated means a future update can remove compile in a later version of gradle).

build.gradle
jsarabia: https://github.com/FibreFoX/javafx-gradle-plugin#minimal-setup-of-buildgradle

build.gradle
danparisd: @mbr4477 Yes, with implementation it was not adding the classpaths to the jar.

@sarabiaj I agree. Gradle says `compile` "is now deprecated — although it won’t be going away any time soon." I did a bit of research and I'm not finding a clear direction on what to do to move away from `compile` for all of our dependencies. I think if dev can give me the config option that is needed for each of the dependencies, that would be best. Since compile isn't going away anytime soon I'm ok with this being made an issue and addressed at a later time.

jsarabia: https://jeroenmols.com/blog/2017/06/14/androidstudio3/

sounds like you could replace compile with api everywhere.

Also sounds like implementation should be fine everywhere except our stuff (common, resource container), and those probably should be api.

build.gradle
danparisd: I removed the jfx plugin and it built and ran, but not sure if there's any particular test that should be run before saying we don't need it.

build.gradle
danparisd: I replaced compile with implementation (except for common and resource container) and it won't run. Specifically it appears to fail with some part of the kotlin jdk.

```PS C:\users\parisd\8woc2018-jvm\build\libs> java.exe -jar 8woc2018-jvm.jar
Exception in thread "main" java.lang.NoClassDefFoundError: kotlin/jvm/internal/Intrinsics
        at org.wycliffeassociates.otter.jvm.app.MainKt.main(Main.kt)
Caused by: java.lang.ClassNotFoundException: kotlin.jvm.internal.Intrinsics
        at java.net.URLClassLoader.findClass(Unknown Source)
        at java.lang.ClassLoader.loadClass(Unknown Source)
        at sun.misc.Launcher$AppClassLoader.loadClass(Unknown Source)
        at java.lang.ClassLoader.loadClass(Unknown Source)
        ... 1 more
```

danparisd: Tried again with api instead of compile and it also doesn't run.

mbr4477: @danparisd Try using `implementation` with this new jar task:
```gradle
//tell gradle what to put in the jar
task fatJar(type: Jar) {
    manifest {
        attributes('Main-Class': mainClassName)
    }
    from {
        configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
    }
    exclude "META-INF/MANIFEST.MF"
    exclude "META-INF/*.SF"
    exclude "META-INF/*.DSA"
    exclude "META-INF/*.RSA"
    with jar
}
```
https://stackoverflow.com/questions/49278063/how-do-i-create-an-executable-fat-jar-with-gradle-with-implementation-dependenci

build.gradle
jsarabia: the jfx plugin is required for the openjdk and for arm architectures. The only reason it builds and runs without it is because javafx is included in the oracle jdk.

I'm not sure if we are allowed to distribute the oracle jre?

mbr4477: I think we can? https://www.oracle.com/technetwork/java/javase/jre-8-readme-2095710.html
> To run your application, a user needs the Java SE Runtime Environment, which is freely available from Oracle. Or, you can redistribute the Java SE Runtime Environment for free with your application, according to the terms of the Oracle Binary Code License Agreement for the Java SE Platform Products.

build.gradle
danparisd: I went with:

```
//tell gradle what to put in the jar
jar {
    manifest {
        attributes(
        'Class-Path': configurations.compile.collect { it.getName() }.join(' '),
        'Main-Class': mainClassName
        )
    }
    from (configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }) {
        exclude "META-INF/MANIFEST.MF"
        exclude "META-INF/*.SF"
        exclude "META-INF/*.DSA"
        exclude "META-INF/*.RSA"
    }
}
```

and it works!

build.gradle
jsarabia: yeah it just may involve including shipping their tos in the installer or something like that.

danparisd:
Closing for new PR with updated changes after comments

PR #317, opened 2018-11-19
Gradle jar resource access

danparisd:
Include Matthews changes for resource access in the JAR. Modify build.gradle for JAR, Install4J, and moving from compile to API and Implementation. Add install4j config file


build.gradle
```diff
@@ -118,11 +126,51 @@ dependencies {
     testCompile 'org.powermock:powermock-api-mockito2:2.+'
     
     //WA Libraries
-    implementation 'org.wycliffeassociates:8woc2018-common'
-    implementation 'org.wycliffeassociates:kotlin-resource-container'
-    implementation 'com.github.WycliffeAssociates:jdenticon-kotlin:-SNAPSHOT'
+    api 'org.wycliffeassociates:8woc2018-common'
+    api 'org.wycliffeassociates:kotlin-resource-container'
+    api 'com.github.WycliffeAssociates:jdenticon-kotlin:-SNAPSHOT'
```
mbr4477, line 131: @danparisd You've done more research on this that I have, but I was under the impression we should use `implementation` everywhere. From the resources you pointed me to, `api` would be useful if we were developing a library and wanted to expose one of our dependencies to consumers in addition to our library's API. However, we are the end user application, no one will be importing us as a library and wanting to use our dependencies, so I think `implementation` would be the way to go here.

build.gradle
```diff
@@ -15,28 +15,36 @@ buildscript {
...
         classpath 'org.xerial:sqlite-jdbc:3.25.2'
+        classpath group: 'com.install4j', name: 'install4j-gradle', version: '7.0.8'
     }
 }
 
 plugins {
     id 'java'
+    id 'java-library'
     id 'nu.studer.jooq' version '3.0.1'
+    id 'application'
```
mbr4477, line 40: Do we end up using the application plugin anywhere other than to get/set the mainClassName variable? If not, it might be better to define that mainClassName up around line 14 and remove our dependency on the plugin.

build.gradle
```diff
@@ -118,11 +126,51 @@ dependencies {
...
 }
 
+//Application plugin setting
+mainClassName = "org.wycliffeassociates.otter.jvm.app.MainKt"
+
+//tell gradle what to put in the jar
+jar {
+    manifest {
+        attributes(
+        'Class-Path': configurations.compile.collect { it.getName() }.join(' '),
```
mbr4477, line 141: Since we are collecting the runtime classpath below, I think we can get rid of this line, especially since we switched from `compile` to `implementation` (which I think means the compile classpath is nearly empty).

build.gradle
danparisd: `If the types from a dependency appear in public fields or methods of your library’s public classes, then that dependency is exposed via your library’s public API and should therefore be added to the api configuration. Otherwise, the dependency is an internal implementation detail and should be added to implementation.` - Gradle Documentation.

I don't know enough about the codebase to speak intelligently to this. Thoughts? It kind of sounds like we might be our own consumer and therefore would need API.

danparisd: Further reference: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies

build.gradle
danparisd: Indeed. Classpath is blank in the manifest.mf inside the jar. Removed it and it compiled exactly the same.

build.gradle
danparisd: I thought that the JAR task depended on the application plugin, but it doesn't. Good catch. I removed the application plugin and just referenced the Main-Class directly in the Jar task and it works.

build.gradle
danparisd: I tested this change and things seem to be working. I thought I tried this already and it wouldn't run, but maybe something we changed elsewhere fixed the issue that I thought was caused by using `implementation` here. If you read through the documentation and think that we're not using any of the things that would need API I'm good with changing to implementation.

mbr4477: **TL;DR: let's go with `implementation` in the jvm repo**

I think the issue here is whether or not we are developing a library. So in this case JVM is not a library, so we can use implementation.

As a contrary example, 8woc2018-common *is* a library (we are depending on it), and it exposes public APIs (e.g., `ImportResourceContainer`) which use data types from RxKotlin. So it would make sense to specify the rx dependencies in common with `api`. The advantage of this is that if I want to use the common library in another project (like the JVM repo) I don't have to add a separate dependency for RxKotlin -- the appropriate RxKotlin version needed for common is already exposed through common itself. If we used `implementation` for the rx dependencies in common, we would be required to have a dependency on rx directly in JVM in order to have the data types needed to interact with the common library. 

This has not raised its head so far since we needed (and so included) RxKotlin for the repo code in JVM anyway, so it hasn't mattered whether the rx api was exposed from common or not.

danparisd: Sounds good.

danparisd:
closing for new PR with discussed changes

build.gradle
```diff
@@ -15,28 +15,36 @@ buildscript {
...
     id 'nu.studer.jooq' version '3.0.1'
+    id 'application'
 }
 
 group 'org.wycliffeassociates'
 apply plugin: 'kotlin'
 apply plugin: 'kotlin-kapt'
-apply plugin: 'edu.sc.seis.launch4j'
+apply plugin: 'install4j'
+//apply plugin: 'javafx-gradle-plugin'
```
jsarabia, line 47: do we no longer need the javafx plugin?

build.gradle
jsarabia: If I'm understanding this correctly, then I think what we *could* do is modify the build.gradle for common, have kotlin-resource-container there use API, and then have everything here be implementation?

Seems to me we landed on api though because class definitions weren't getting packaged into the jar though.

build.gradle
danparisd: Not that I can tell since we are using the Oracle JRE.

build.gradle
danparisd: I think they are getting packaged even with implementation due to the `runtimeClasspath` change, now.

PR #332, opened 2018-11-29
Filter source language from target combo box

mbr4477:
- Removes selected source language from target combo box so the user cannot select the same language for source and target


mbr4477:
Getting rid of combo boxes

PR #335, opened 2018-11-29
Use DB keystore instead of Preferences API

mbr4477:


jsarabia (changes requested):
the convenience methods are nice.
This is still a database access though, so we should still likely be following the pattern of doing this all with rx

jsarabia (approved):

PR #337, opened 2018-12-03
Add onErrorComplete to init selected

mbr4477:
- keeps exception from being thrown if no plugins exist (causing completable to get a no element exception)


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/AudioPluginRepository.kt
```diff
@@ -107,6 +107,7 @@ class AudioPluginRepository(
                             return@flatMapCompletable Completable.complete()
                         }
                     }
+                    .onErrorComplete() // complete even if no plugins to initialize
```
jsarabia: Can't we handle this in the flatmap rather than silence all potential errors?

jsarabia (approved):

PR #361, opened 2018-12-18
Rc uniqueness updates

mbr4477:


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/CollectionRepository.kt
```diff
@@ -186,25 +193,30 @@ class CollectionRepository(
...
+            }
+            creator = "otter"
+            version = metadata.version
+            format = "text/usfm"
+            subject = metadata.subject
+            type = "book"
+            title = metadata.title
+        }
+        val directory = directoryProvider.getDerivedContainerDirectory(
+                dublinCore.mapToMetadata(File("."), targetLanguage),
```
jsarabia, line 213: What is File(".") supposed to resolve to?

mbr4477: Needed a placeholder. I'm trying to generate the folder for an RC being derived from existing content. To do that I need to figure out what folder the new RC should go in. The `getDerivedContainerDirectory()` function is how we do that, but it requires metadata to get things like creator, identifier, etc. Since I have dublin core data, I need to use mapToMetadata to get the metadata. The map to metadata function requires a folder, but in this case I need to get the metadata first in order to find out what that folder name should be. So the metadata has a placeholder directory in it, but that placeholder doesn't get used by the `getDerivedContainerDirectory` function. 

I think I stuck with requiring metadata instead of dublin core in the DP API since that seemed to best to keep DP from being tied closely to the RC library

jsarabia (approved):

PR #399, opened 2019-02-05
Added 'end' to Content and 'rc_fk' to Resource

KJoslyn:
`end` was a field in the `Content` model object (Content.kt in the common repo), but it had not been added to the Content table. This PR accomplishes this.

Additionally, I added `rc_fk` to the Resource (`resource_link`) table. This addition will allow us to retrieve specific types of resources (tn, tq, etc).

In ResourceRepository.kt, I had to add an `rcFk` parameter to several functions since the `Collection` object's `resourceContainer` field is nullable, but `rc_fk ` is not nullable in the Resource table. This doesn't feel very good. Why do we allow `Collection`'s `resourceContainer` field to be null?

This brings up another point. Should `rc_fk` in fact be non-nullable? I think so, since we are not allowing users to create their own notes (if we did someday though, these would not be tied to a resource container.)

This code will not build until the corresponding PR #[83](https://github.com/WycliffeAssociates/otter-common/pull/83) in the common repo is merged.

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentDao.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentDao.kt
@@ -67,6 +67,7 @@ class ContentDao(
                         CONTENT_ENTITY.COLLECTION_FK,
                         CONTENT_ENTITY.SORT,
                         CONTENT_ENTITY.START,
+                        CONTENT_ENTITY.END,
                         CONTENT_ENTITY.LABEL,
                         CONTENT_ENTITY.SELECTED_TAKE_FK,
                         CONTENT_ENTITY.TEXT,
@@ -76,6 +77,7 @@ class ContentDao(
                         entity.collectionFk,
                         entity.sort,
                         entity.start,
+                        entity.end,
                         entity.labelKey,
                         entity.selectedTakeFk,
                         entity.text,
@@ -117,6 +119,7 @@ class ContentDao(
                 .set(CONTENT_ENTITY.SORT, entity.sort)
                 .set(CONTENT_ENTITY.LABEL, entity.labelKey)
                 .set(CONTENT_ENTITY.START, entity.start)
+                .set(CONTENT_ENTITY.END, entity.end)
                 .set(CONTENT_ENTITY.COLLECTION_FK, entity.collectionFk)
                 .set(CONTENT_ENTITY.SELECTED_TAKE_FK, entity.selectedTakeFk)
                 .set(CONTENT_ENTITY.TEXT, entity.text)
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/entities/ResourceLinkEntity.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/entities/ResourceLinkEntity.kt
@@ -5,5 +5,5 @@ data class ResourceLinkEntity(
         var resourceContentFk: Int,
         var contentFk: Int?,
         var collectionFk: Int?,
-        var rc_fk: Int
+        var rcFk: Int
 )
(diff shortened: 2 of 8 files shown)


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/entities/ResourceLinkEntity.kt
```diff
@@ -4,5 +4,6 @@ data class ResourceLinkEntity(
         var id: Int,
         var resourceContentFk: Int,
         var contentFk: Int?,
-        var collectionFk: Int?
+        var collectionFk: Int?,
+        var rc_fk: Int
```
aunger: In the entities package, convention seems to be camelCase, as opposed to underscore naming in the DB.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentDao.kt
```diff
@@ -67,6 +67,7 @@ class ContentDao(
                         CONTENT_ENTITY.COLLECTION_FK,
                         CONTENT_ENTITY.SORT,
                         CONTENT_ENTITY.START,
+                        CONTENT_ENTITY.END,
```
aunger, line 70: Do we need `end`?  I know it already exists in some places, but maybe it'd work to delete those instead.  It will be extra work to create it at import time in some cases, and I'm wondering what benefit we'll get from it being in the db, rather than just calculating it if and when it's ever needed.  

I'm sure @sarabiaj has thought this through. Joe, was this omitted as an oversight, or on purpose?

aunger (changes requested):
~~Does this require changes in the [other repo](https://github.com/WycliffeAssociates/otter-common)? I don't see a PR there yet.~~ Resolved, see [common#83](https://github.com/WycliffeAssociates/otter-common/pull/83)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceRepository.kt
```diff
@@ -68,7 +68,7 @@ class ResourceRepository(
                 .subscribeOn(Schedulers.io())
     }
 
-    override fun linkToContent(resource: Content, content: Content): Completable {
+    override fun linkToContent(resource: Content, content: Content, rcFk: Int): Completable {
```
KJoslyn, line 71: I had to add an `rcFk` parameter to this function and the one below since the `Collection` object's `resourceContainer` field is nullable, but `rc_fk` is not nullable in the Resource table.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/entities/ResourceLinkEntity.kt
KJoslyn: > Does this require changes in the [other repo](https://github.com/WycliffeAssociates/otter-common)? I don't see a PR there yet.

It did, I missed something. It is there now. [#83](https://github.com/WycliffeAssociates/otter-common/pull/83). The `Content` data model object already had the `end` field, and we don't yet have a `Resource` data model object, so there was only one file that needed changing and I didn't catch it.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentDao.kt
KJoslyn: I see that in your `MarkdownProjectReader` implementation, you currently set both `start` and `end` to the `fileId`, which apparently will be 0 for the intro.md files. We had talked about using both `start` and `end` before, but I don't think there's a reason to if the only entries that will use it are contents that are resources for chapters (`start` = 0 seems to work just fine.)

aunger: Yes, that's an example of where it'd make the code messier to try to precalculate an end. There may be a very important use-case for this field re chunked scripture. However, I propose that we defer adding `end` if we don't need it right now. (I'm wondering whether even that chunk use-case would be better served by calculating `end` as needed.)

jsarabia: I had omitted it because it could be derived, and could just be calculated on mapping entities to our app data classes; but it's also probably not a big deal and if you really want to add it, it really wouldn't be much of a cost.

PR #400, opened 2019-02-07
Added rc_fk to Resource table

KJoslyn:
Added resource container foreign key to the Resource table so we can query the database for resources of a particular type (tn, tq, etc).

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/RecordMappers.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/RecordMappers.kt
@@ -76,7 +76,8 @@ class RecordMappers {
                     record.getValue(RESOURCE_LINK.ID),
                     record.getValue(RESOURCE_LINK.RESOURCE_CONTENT_FK),
                     record.getValue(RESOURCE_LINK.CONTENT_FK),
-                    record.getValue(RESOURCE_LINK.COLLECTION_FK)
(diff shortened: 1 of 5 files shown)


jsarabia (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/RecordMappers.kt
```diff
@@ -76,7 +76,8 @@ class RecordMappers {
                     record.getValue(RESOURCE_LINK.ID),
                     record.getValue(RESOURCE_LINK.RESOURCE_CONTENT_FK),
                     record.getValue(RESOURCE_LINK.CONTENT_FK),
-                    record.getValue(RESOURCE_LINK.COLLECTION_FK)
+                    record.getValue(RESOURCE_LINK.COLLECTION_FK),
+                    record.getValue(RESOURCE_LINK.RC_FK)
```
aunger, line 80: I hate to do this right now.  I know we've already discussed this.  But...

Would it be better for this to be in the Content table?  That would mean that selecting all the content for a given location and RC would use the same process whether it's a resource or main text.  Done this way (with rcfk in the resource table), to filter by RC, you'd join with that table in the case of resources, but join with parent collection in the case of main text.

If the answer is, we've thought this through already, that's fine.

jsarabia: I might have to come back to this tomorrow, but I am willing to revisit it

Issue #369, opened 2019-01-22
Mapping Markdown tags to database records


Redmayspecial:
## User Story
As a manager, I need a process to take all the mapped data from the parser process and write it to the database.

PR #403, opened 2019-02-07
RcTreeImporter

KJoslyn:
This PR is for issue #[369](https://github.com/WycliffeAssociates/otter-jvm/issues/369), importing a resource container tree that may either be a scripture resource container or a "help" resource container. Please see the corresponding PR in common #[85](https://github.com/WycliffeAssociates/otter-common/pull/85).


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/RCImporter.kt
```diff
@@ -0,0 +1,124 @@
...
+import org.wycliffeassociates.otter.common.persistence.repositories.IContentRepository
+import org.wycliffeassociates.otter.jvm.persistence.database.AppDatabase
+import org.wycliffeassociates.otter.jvm.persistence.entities.ResourceLinkEntity
+import org.wycliffeassociates.otter.jvm.persistence.repositories.mapping.CollectionMapper
+import org.wycliffeassociates.otter.jvm.persistence.repositories.mapping.ContentMapper
+import org.wycliffeassociates.otter.jvm.persistence.repositories.mapping.LanguageMapper
+import org.wycliffeassociates.otter.jvm.persistence.repositories.mapping.ResourceMetadataMapper
+import org.wycliffeassociates.resourcecontainer.ResourceContainer
+
+class RcImporter(
```
aunger: Filename case should match.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/RcTreeImporter.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    override fun importResourceContainer(rc: ResourceContainer, rcTree: Tree, languageSlug: String): Completable {
+        return Completable.fromAction {
+            database.transaction { dsl ->
+                val language = LanguageMapper().mapFromEntity(database.getLanguageDao().fetchBySlug(languageSlug, dsl))
+                val helpRcTargetMetadata: ResourceMetadata? =
+                        when (rc.type()) {
+                            // TODO: Identifier shouldn't just be set to "ult"
+                            "help" -> ResourceMetadataMapper().mapFromEntity(
+                                    database.getResourceMetadataDao().fetchByIdentifier("ult", dsl),
```
KJoslyn: @sarabiaj how should I go about looking for the corresponding scripture resource container? Could this be what the "RC Links" table is for? Currently in the manifest for the ULT I see there is a "relation" field with "en/tn", "en/tq", and "en/tw". Likewise, in the manifest for en/tn, there is a "relation" field with "en/ulb" and "en/udb". (In fact, does ulb even exist anymore- I believe it got renamed to ult?) If this is not what "RC Links" is for, do we need a field in the dublin core table for relations?

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/RcTreeImporter.kt
```diff
@@ -0,0 +1,124 @@
...
+                for (childNode in node.children) {
+                    importNode(id, childNode)
+                }
+            }
+        }
+
+        private fun findCollectionId(collection: Collection, helpRcTargetMetadata: ResourceMetadata): Int {
+            // TODO: What to do instead of blocking get?
+            return collectionRepository.getBySlugAndContainer(collection.slug, helpRcTargetMetadata)
+                    .blockingGet()
```
KJoslyn: I am fairly sure we don't want to just do a blockingGet() without catching an error (no such collection). I intend to change this

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/RcTreeImporter.kt
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        private fun linkResource(parentId: Int, contentId: Int, start: Int) {
+            var collectionFk: Int? = null
+            var contentFk: Int? = null
+            when (start) {
+                0 -> collectionFk = parentId
+                // TODO: What to do instead of blocking get?
+                else -> contentFk = contentRepository.getByCollectionIdAndStart(parentId, start)
+                        .blockingGet()
```
KJoslyn: Same as above.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/RcTreeImporter.kt
```diff
@@ -0,0 +1,124 @@
...
+                val metadata = rc.manifest.dublinCore.mapToMetadata(rc.dir, language)
+                val dublinCoreFk = database.getResourceMetadataDao().insert(ResourceMetadataMapper().mapToEntity(metadata), dsl)
+                val ih = ImportHelper(dublinCoreFk, helpRcTargetMetadata, dsl)
+
+                ih.importCollection(null, rcTree)
+            }
+        }.subscribeOn(Schedulers.io())
+    }
+
+    inner class ImportHelper(val dublinCoreId: Int, val helpRcTargetMetadata: ResourceMetadata?, val dsl: DSLContext) {
```
KJoslyn: I could have created a data class with these values as well, but I didn't want to have to pass around all of these static values to every function.  Since I instantiate the RcTreeImporter in Injector.kt, I could not have these simply be member variables of the RcTreeImporter class itself.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/RcTreeImporter.kt
aunger: Consider this: 
```
        fun importCollection(parentId: Int?, node: Tree) =
                (node.value as? Collection)?.let { collection ->
                    when (helpRcTargetMetadata) {
                        null -> addCollection(collection, parentId)
                        else -> findCollectionId(collection, helpRcTargetMetadata)
                    }
                }?.subscribe { id ->
                    for (childNode in node.children) {
                        importNode(id, childNode)
                    }
                }

        private fun findCollectionId(collection: Collection, helpRcTargetMetadata: ResourceMetadata): Maybe<Int> =
                collectionRepository
                        .getBySlugAndContainer(collection.slug, helpRcTargetMetadata)
                        .map(Collection::id)

        private fun addCollection(collection: Collection, parentId: Int?): Maybe<Int> =
                Maybe.fromCallable {
                    val entity = CollectionMapper().mapToEntity(collection).apply {
                        parentFk = parentId
                        dublinCoreFk = dublinCoreId
                    }
                    database.getCollectionDao().insert(entity, dsl)
                }
```

aunger (changes requested):

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/RcTreeImporter.kt
KJoslyn: @aunger To keep you in the loop, I talked to Joe about this earlier, and this is indeed what the "RC Links" table is for. Until this morning I hadn't realized that "help" resource containers can be linked to multiple "bundle" RC's (such as ULB and UDB). We decided that for now I will duplicate the content and resource entries that I am adding to the database for the notes. This is so that each entry can be linked (via the content "belongsTo" field) to the corresponding collection in either the ULB or UDB. Joe said that we are not currently using the UDB so this is a simple solution for now. Currently, I am working on the code to write entries to the "RC Links" table and to handle the general case of a variable number related resource containers.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ResourceMetadataDao.kt
```diff
@@ -115,6 +115,29 @@ class ResourceMetadataDao(
...
+                        .on(DUBLIN_CORE_ENTITY.LANGUAGE_FK.eq(LANGUAGE_ENTITY.ID)))
+                .where(LANGUAGE_ENTITY.SLUG.eq(languageSlug))
+                .and(DUBLIN_CORE_ENTITY.IDENTIFIER.eq(identifier))
+                .fetchOne()
+                ?.let {
+                    RecordMappers.mapToResourceMetadataEntity(it)
+                }
+    }
+
+    fun fetchByIdentifier(identifier: String, dsl: DSLContext = instanceDsl): ResourceMetadataEntity {
```
jsarabia: this isn't sufficient to retrieve a unique metadata entry; for example, all GLs will have a "ulb," so if you import the French ulb and the English ulb, you can't be sure that the fetchOne call will give you the one you want

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ResourceMetadataDao.kt
```diff
@@ -115,6 +115,29 @@ class ResourceMetadataDao(
                 }
     }
 
+    fun fetchByLanguageAndIdentifier(languageSlug: String, identifier: String, dsl: DSLContext = instanceDsl): ResourceMetadataEntity? {
```
jsarabia: This.... also might not be enough to get a unique metadata entry. I don't remember exactly what we decided the unique constraint to be, but version may come into play as well.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+                }
+            }
+        }.subscribeOn(Schedulers.io())
+    }
+
+    private fun linkRelatedResourceContainers(
+            dublinCoreFk: Int,
+            relations: List<String>,
+            dsl: DSLContext)
+            : List<Int> {
```
jsarabia:
```
    dsl: DSLContext
) : List<Int> {
```

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+                    ?.let { relatedDublinCore ->
+                        // TODO: Only add link if it doesn't exist already
+                        resourceMetadataDao.addLink(dublinCoreFk, relatedDublinCore.id, dsl)
+                        relatedIds.add(relatedDublinCore.id)
+                    }
+        }
+        return relatedIds
+    }
+
+    inner class ImportHelper(val dublinCoreId: Int, val relatedBundleDublinCoreId: Int?, val dsl: DSLContext, var collectionsImported: Int = 0) {
```
jsarabia:
```
inner class ImportHelper(
    val dublinCoreId: Int,
    val relatedBundleDublinCoreId: Int?, 
    val dsl: DSLContext, 
    var collectionsImported: Int = 0
) {
```

(one reason for line length limits is that it keeps you from having to scroll horizontally when looking at code on github lol)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+                // TODO: If we don't find a corresponding collection, we continue on, passing null to collectionId.
+                // TODO ... Eventually, contents will not be created if there is no parentId. This will happen for front
+                // TODO ... matter until we have another solution.
+                for (childNode in node.children) {
+                    importNode(collectionId, childNode)
+                }
+            }
+            collectionsImported++
+            if (collectionsImported % 25 == 0) {
+                println("Collections imported: $collectionsImported")
```
jsarabia: unsure if we should be committing print statements like this

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+
+        private fun findCollectionId(collection: Collection, containerId: Int): Int? =
+                collectionDao.fetchBySlugAndContainerId(collection.slug, containerId)
+                        ?.id
```
jsarabia: i'd just put the ?.id on the previous line

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+                when (relatedBundleDublinCoreId) {
+                    null -> addCollection(collection, parentId)
+                    else -> findCollectionId(collection, relatedBundleDublinCoreId)
+                }
+            }.let { collectionId ->
+                // TODO: If we don't find a corresponding collection, we continue on, passing null to collectionId.
+                // TODO ... Eventually, contents will not be created if there is no parentId. This will happen for front
+                // TODO ... matter until we have another solution.
+                for (childNode in node.children) {
+                    importNode(collectionId, childNode)
```
jsarabia: I can't think of a situation where the types don't match (like, the children are either all content, or all collections), so I suppose one thing we could do is rather than iterate and import nodes one at a time, you could find the type here, and insert all into a list, and then insert the list into the database at once.

If the assumption doesn't hold (and they could be mixed types), then you could filter them out into two lists.

I think we'd probably have other issues if they did mix and match anyway though?

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
KJoslyn: Yeah for the time being since the import takes so long it's the only indication that I have that it's still running, but it wasn't supposed to make it into the PR!

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentDao.kt
```diff
@@ -20,6 +20,17 @@ class ContentDao(
                 }
     }
 
+    fun fetchVerseByCollectionIdAndStart(collectionId: Int, start: Int, dsl: DSLContext = instanceDsl): ContentEntity? {
+        return dsl
+                .select()
+                .from(CONTENT_ENTITY)
+                .where(CONTENT_ENTITY.COLLECTION_FK.eq(collectionId))
+                .and(CONTENT_ENTITY.START.eq(start))
+                .and(CONTENT_ENTITY.LABEL.eq("verse")) // TODO: Probably shouldn't be hardcoded
```
jsarabia, line 29: I wouldn't go by the label.

For example, OBS has TN as well. Understanding we currently cannot parse OBS to import it anyway, if we did, it has chunks, not verses.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+                        resourceMetadataDao.addLink(dublinCoreFk, relatedDublinCore.id, dsl)
+                        relatedIds.add(relatedDublinCore.id)
+                    }
+        }
+        return relatedIds
+    }
+
+    inner class ImportHelper(val dublinCoreId: Int, val relatedBundleDublinCoreId: Int?, val dsl: DSLContext, var collectionsImported: Int = 0) {
+
+        fun importCollection(parentId: Int?, node: Tree) {
```
jsarabia: this should be moved further down, to be near importNode, importContent (to keep related things organized, close to where they're used, etc)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentDao.kt
aunger: Re todo: I'd prefer to have a `fetchContentByRcAndCollectionAndStart` filtered on RC_FK instead of label.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
jsarabia: It's also possible this wouldn't affect anything anyway because jooq might hold off on inserting?

jsarabia: no actually, it probably can't wait, and the insert is returning an id right away

I forgot- trying to optimize this stuff is probably going to mean doing things down inside the dao level, not at a high level like this

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentDao.kt
KJoslyn: Unfortunately I'm not sure how to handle this right now, because we currently are putting a Content entry into the DB that corresponds to a chapter, and which has a start of "0". For now, label is the only differentiating factor (see the image of the Content table below.) That content entry corresponding to the chapter was being placed there in the previous version of the importer, and I'm not sure what purpose it has. Perhaps it shouldn't be placed there in the first place. Is it ok if we create a story to address these issues and leave it the way it is now for the PR? (It's ok if the answer is no.)
image

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ResourceMetadataDao.kt
aunger: [UNIQUE (language_fk, identifier, version, creator, derivedFrom_fk)](https://github.com/WycliffeAssociates/otter-jvm/blob/84520bd367b98259643e2b0932a9bb2c019c6dcb/src/main/resources/sql/CreateAppDb.sql#L27)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
KJoslyn: That might work- perhaps we can create a story to accomplish batch inserting later? (By later I mean soon)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ResourceMetadataDao.kt
KJoslyn: This function isn't being used anyway, I should've already removed it.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+                val language = LanguageMapper().mapFromEntity(languageDao.fetchBySlug(languageSlug, dsl))
+                val metadata = rc.manifest.dublinCore.mapToMetadata(rc.dir, language)
+                val dublinCoreFk = resourceMetadataDao.insert(ResourceMetadataMapper().mapToEntity(metadata), dsl)
+
+                // TODO: Should this only happen if we are importing a "help" RC? What if the help is imported first,
+                // TODO: ... then another related bundle RC is imported?
+                val relatedDublinCoreIds: List<Int> =
+                        linkRelatedResourceContainers(dublinCoreFk, rc.manifest.dublinCore.relation, dsl)
+
+                // TODO: Probably shouldn't hardcode "help"
```
aunger: Re todo: I think the RC standard forces this. There's nothing in the manifest's `relation` that tells which resource is primary. tN links to ulb, and ulb links to tN.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ResourceMetadataDao.kt
KJoslyn: Hm. I know we had talked about version previously and the fact that the version numbers might not match up between related resource containers? Really not sure how to handle this for now though.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
aunger: I agree, but would look the other way here, given that an import job currently takes close to an hour.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+        private fun addCollection(collection: Collection, parentId: Int?): Int {
+            val entity = CollectionMapper().mapToEntity(collection).apply {
+                parentFk = parentId
+                dublinCoreFk = dublinCoreId
+            }
+            return collectionDao.insert(entity, dsl)
+        }
+
+        private fun importNode(parentId: Int?, node: TreeNode) {
+            when (node) {
```
aunger: I've been _almost_ commenting on this line for days.  I can't rationally justify it, but I can't shake the feeling that this `when` should check the type of `node.value` for Collection vs Content instead.

aunger:
Drat, closing this via ZenHub was a mistake.  Reopening.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
aunger: Joe corrected me. The manifest's `type: 'help'` has a special meaning in the RC spec, indicating a secondary collection. https://resource-container.readthedocs.io/en/latest/container_types.html#help-help

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ContentRepository.kt
```diff
@@ -1,6 +1,7 @@
 package org.wycliffeassociates.otter.jvm.persistence.repositories
 
 import io.reactivex.Completable
+import io.reactivex.Maybe
```
natibekele: are you using this?

aunger (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ContentRepository.kt
KJoslyn: Just removed it.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ResourceMetadataDao.kt
```diff
@@ -115,6 +115,27 @@ class ResourceMetadataDao(
...
+                            dsl: DSLContext = instanceDsl
+    ) : ResourceMetadataEntity? {
+        return dsl
+                .select()
+                .from(DUBLIN_CORE_ENTITY.join(LANGUAGE_ENTITY)
+                        .on(DUBLIN_CORE_ENTITY.LANGUAGE_FK.eq(LANGUAGE_ENTITY.ID)))
+                .where(LANGUAGE_ENTITY.SLUG.eq(languageSlug))
+                .and(DUBLIN_CORE_ENTITY.IDENTIFIER.eq(identifier))
+                .and(DUBLIN_CORE_ENTITY.CREATOR.eq(creator))
+                .and(DUBLIN_CORE_ENTITY.DERIVEDFROM_FK.eq(derivedFromFk))
```
aunger:
```
                .and(derivedFromFk?.let(DUBLIN_CORE_ENTITY.DERIVEDFROM_FK::eq)
                        ?: DUBLIN_CORE_ENTITY.DERIVEDFROM_FK.isNull)
```

aunger (changes requested):

aunger (approved):

Issue #380, opened 2019-01-23
Display verse/chunk cards in a resource tab


KJoslyn:
As a narrator, when I am in a resource tab, I would like to see a card for each verse/chunk/chapter so that I can see all of the translationNotes attached to each card.

Redmayspecial:
![image](https://user-images.githubusercontent.com/11600863/51618528-7c72c900-1efc-11e9-8ac2-1913b40496b6.png)

aunger:
Connected to task #382

PR #435, opened 2019-03-13
tn cards 'n tabs

aunger:
This is just a PR for @KJoslyn -- proposed additions to his branch.  This will bring you a little closer to what we talked about today.  The main update is using `Property<ResourceMetadata>` instead of `<String>` to track which tab is selected.  I think that'll make it easier to remove hard-coded enum values, etc.  (There's also some progress toward #380 that's hard to split out from these changes, so bonus.)

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/collectionsgrid/viewmodel/CollectionsGridViewModel.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/collectionsgrid/viewmodel/CollectionsGridViewModel.kt
@@ -2,26 +2,21 @@ package org.wycliffeassociates.otter.jvm.app.ui.collectionsgrid.viewmodel
 
 import com.github.thomasnield.rxkotlinfx.observeOnFx
 import com.github.thomasnield.rxkotlinfx.toObservable
-import javafx.beans.property.SimpleBooleanProperty
-import javafx.beans.property.SimpleObjectProperty
 import javafx.collections.FXCollections
 import javafx.collections.ObservableList
 import org.wycliffeassociates.otter.common.data.model.Collection
-import org.wycliffeassociates.otter.common.data.model.Content
-import org.wycliffeassociates.otter.common.domain.content.AccessTakes
-import org.wycliffeassociates.otter.jvm.app.ui.collectionsgrid.view.CollectionGridStyles
+import org.wycliffeassociates.otter.common.data.model.ResourceMetadata
 import org.wycliffeassociates.otter.jvm.app.ui.collectionsgrid.view.CollectionTabPane
 import org.wycliffeassociates.otter.jvm.app.ui.inject.Injector
 import org.wycliffeassociates.otter.jvm.app.ui.mainscreen.view.MainScreenStyles
-import tornadofx.*
+import tornadofx.CssRule
+import tornadofx.ViewModel
+import tornadofx.getProperty
+import tornadofx.property
 
 class CollectionsGridViewModel : ViewModel() {
     private val injector: Injector by inject()
-    private val directoryProvider = injector.directoryProvider
     private val collectionRepository = injector.collectionRepo
-    private val contentRepository = injector.contentRepository
-    private val takeRepository = injector.takeRepository
-    private val pluginRepository = injector.pluginRepository
     private val resourceMetadataRepository = injector.resourceMetadataRepository
 
     // The selected project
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/collectionsgrid/viewmodel/CollectionsGridViewModel.kt
```diff
@@ -2,26 +2,21 @@ package org.wycliffeassociates.otter.jvm.app.ui.collectionsgrid.viewmodel
...
 import javafx.collections.ObservableList
 import org.wycliffeassociates.otter.common.data.model.Collection
-import org.wycliffeassociates.otter.common.data.model.Content
-import org.wycliffeassociates.otter.common.domain.content.AccessTakes
-import org.wycliffeassociates.otter.jvm.app.ui.collectionsgrid.view.CollectionGridStyles
+import org.wycliffeassociates.otter.common.data.model.ResourceMetadata
 import org.wycliffeassociates.otter.jvm.app.ui.collectionsgrid.view.CollectionTabPane
 import org.wycliffeassociates.otter.jvm.app.ui.inject.Injector
 import org.wycliffeassociates.otter.jvm.app.ui.mainscreen.view.MainScreenStyles
-import tornadofx.*
```
KJoslyn, line 16: According to the tornadofx documentation we should be importing `tornadofx.*` in all tornadofx-related files https://edvin.gitbooks.io/tornadofx-guide/content/part1/3.%20Components.html

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/mainscreen/viewmodel/MainViewViewModel.kt
```diff
@@ -49,29 +58,34 @@ class MainViewViewModel : ViewModel() {
...
             activeCollection.bindBidirectional(selectedCollectionProperty)
+            activeResource.bindBidirectional(selectedResourceProperty)
             activeContent.bindBidirectional(selectedContentProperty)
-
         }
     }
 
-    fun contentSelected(content: Content) {
+    private fun resourceSelected(resource: ResourceMetadata) {
+        // TODO: refilter collections and change colors in CollectionGrid here
```
KJoslyn, line 85: Why wouldn't we just do this in `CollectionsGridViewModel`?

...Though for styling, I realize that the color scheme changes even for the content grid and the takes page. Since I was anticipating using the same css class for the elements on those pages as well, we could have another property to go along with `activeProject`, `activeCollection` etc., called `activeResourceStyleClass` or something like that. (I already have `selectedResourceClassProperty` in `CollectionsGridViewModel`.) I'm not sure how I feel about a `cssRule` property existing alongside the other "active" properties, but at least we wouldn't have to duplicate code in all of the view models to accomplish the same thing.

aunger: We probably would (and I think I already do). Sorry for dumping half-baked code on you.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/contentgrid/view/ContentGrid.kt
```diff
@@ -2,32 +2,84 @@ package org.wycliffeassociates.otter.jvm.app.ui.contentgrid.view
...
+            createBooleanBinding(Callable { activeResource.value.type == "help" }, activeResource)
+
     init {
         importStylesheet<ContentGridStyles>()
         importStylesheet<DefaultStyles>()
         importStylesheet<MainScreenStyles>()
     }
 
+    private fun buildContentWidget(contentGroup: GroupedContent): Node {
+        return when (activeResource.value?.type) {
```
KJoslyn, line 44: I guess you didn't want to check `activeResourceIsHelp` here instead because there will be other types later? (Like tw, which is 'dict')

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/collectionsgrid/viewmodel/CollectionsGridViewModel.kt
aunger: Oh, I think that's so auto-complete can find tfx extension methods on non-tfx classes. This should be changed back.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/contentgrid/view/ContentGrid.kt
aunger: Because I hadn't added isHelp yet when I wrote this line. But your explanation sounds good.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/mainscreen/viewmodel/MainViewViewModel.kt
KJoslyn: @aunger I edited my comment about this. I think I'll go ahead and merge the PR into my branch- perhaps this todo message can just be changed to something like "Consider setting a shared `cssRule` property here"

PR #440, opened 2019-04-03
ResourceCard, ResourceGroupCard, and HighlightableButton widgets

KJoslyn:
Includes resource card, resource group card, and highlightable button widgets.

Requires common [#96](https://github.com/WycliffeAssociates/otter-common/pull/96).

To test, check out [kj-resources-testapp](https://github.com/WycliffeAssociates/otter-jvm/tree/kj-resources-testapp) and run `main` inside `ResourceCardApp.kt`. (This is a second entry point that is not included in this PR.)

In `ResourceCard.kt`, I plan to incorporate Nathan's status indicators very soon. Could be part of this PR but maybe the next one.

ViewModels and Views (other than widgets) are not included in this PR.

NEW 4/5/19: To test the `primaryColor` property of the ResourceCard, go to `ResourceGroupCard.kt` and change line 15 to
```
resourcecard(it) {
     primaryColor = Color.(X)
}
```

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
@@ -4,13 +4,12 @@ import com.jfoenix.controls.JFXButton
 import de.jensd.fx.glyphs.materialicons.MaterialIconView
 import javafx.beans.property.SimpleBooleanProperty
 import javafx.scene.paint.Color
-import org.wycliffeassociates.otter.jvm.app.theme.AppTheme
 import tornadofx.*
 
 class HighlightableButton : JFXButton() {
 
-    var primaryColor: Color = AppTheme.colors.black
-    var secondaryColor: Color = AppTheme.colors.white
+    var primaryColor: Color = Color.BLACK
+    var secondaryColor: Color = Color.WHITE
     var hasBorder: Boolean = false
     var isHighlighted: SimpleBooleanProperty = SimpleBooleanProperty(false)
 
@@ -32,13 +31,13 @@ class HighlightableButton : JFXButton() {
 
     fun applyColors() {
         if (hoverProperty().get() || isHighlighted.get()) {
-            applyColors(primaryColor, secondaryColor)
+            doApplyColors(primaryColor, secondaryColor)
         } else {
-            applyColors(secondaryColor, primaryColor)
+            doApplyColors(secondaryColor, primaryColor)
         }
     }
 
-    private fun applyColors(bgColor: Color, contentFillColor: Color) {
+    private fun doApplyColors(bgColor: Color, contentFillColor: Color) {
         style {
             backgroundColor += bgColor
             if(hasBorder) {
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
@@ -3,15 +3,17 @@ package org.wycliffeassociates.otter.jvm.app.widgets.highlightablebutton
 import com.jfoenix.controls.JFXButton
 import de.jensd.fx.glyphs.materialicons.MaterialIconView
 import javafx.beans.property.SimpleBooleanProperty
+import javafx.beans.property.SimpleObjectProperty
 import javafx.scene.paint.Color
 import tornadofx.*
 
 class HighlightableButton : JFXButton() {
 
-    var primaryColor: Color = Color.BLACK
+    var isHighlightedProperty: SimpleBooleanProperty = SimpleBooleanProperty(false)
+    var highlightColorProperty = SimpleObjectProperty<Color>(Color.BLACK)
+    var highlightColor: Color by highlightColorProperty
     var secondaryColor: Color = Color.WHITE
     var hasBorder: Boolean = false
-    var isHighlighted: SimpleBooleanProperty = SimpleBooleanProperty(false)
 
     init {
         importStylesheet<HighlightableButtonStyles>()
@@ -19,21 +21,16 @@ class HighlightableButton : JFXButton() {
         addClass(HighlightableButtonStyles.hButton)
         isDisableVisualFocus = true
 
-        isHighlighted.onChange {
-            applyColors()
-        }
-        hoverProperty().onChange {
-            applyColors()
-        }
-
-        applyColors()
+        isHighlightedProperty.onChange { applyColors() }
+        hoverProperty().onChange { applyColors() }
+        highlightColorProperty.onChange { applyColors() }
     }
 
     fun applyColors() {
-        if (hoverProperty().get() || isHighlighted.get()) {
-            doApplyColors(primaryColor, secondaryColor)
+        if (hoverProperty().get() || isHighlightedProperty.get()) {
+            doApplyColors(highlightColor, secondaryColor)
         } else {
-            doApplyColors(secondaryColor, primaryColor)
+            doApplyColors(secondaryColor, highlightColor)
         }
     }
 
@@ -41,7 +38,7 @@ class HighlightableButton : JFXButton() {
         style {
             backgroundColor += bgColor
             if(hasBorder) {
-                borderColor += box(primaryColor)
+                borderColor += box(highlightColor)
             }
         }
         (graphic as? MaterialIconView)?.apply {
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
@@ -4,33 +4,33 @@ import com.jfoenix.controls.JFXButton
 import de.jensd.fx.glyphs.materialicons.MaterialIconView
 import javafx.beans.property.SimpleBooleanProperty
 import javafx.scene.paint.Color
+import org.wycliffeassociates.otter.jvm.app.theme.AppTheme
 import tornadofx.*
 
-class HighlightableButton(
-        private val primaryColor: Color,
-        private val secondaryColor: Color,
-        private val hasBorder: Boolean,
-        private val isHighlighted: SimpleBooleanProperty,
-        private val icon: MaterialIconView? = null
-) : JFXButton() {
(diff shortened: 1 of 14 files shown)


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
```diff
@@ -0,0 +1,53 @@
...
+import javafx.beans.property.SimpleBooleanProperty
+import javafx.scene.paint.Color
+import tornadofx.*
+
+class HighlightableButton(
+        private val primaryColor: Color,
+        private val secondaryColor: Color,
+        private val hasBorder: Boolean,
+        private val isHighlighted: SimpleBooleanProperty,
+        private val icon: MaterialIconView? = null
```
KJoslyn: I didn't want to put this here, but the fill has to be applied directly on the MaterialIconView (applying the fill to the graphic Node does not work, and I do not know how to access the MaterialIconView through the `graphic` field.) All of our use cases have icons, but I made this a nullable parameter.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.jfoenix.controls.JFXButton
+import de.jensd.fx.glyphs.materialicons.MaterialIconView
+import javafx.beans.property.SimpleBooleanProperty
+import javafx.scene.paint.Color
+import tornadofx.*
+
+class HighlightableButton(
+        private val primaryColor: Color,
+        private val secondaryColor: Color,
+        private val hasBorder: Boolean,
```
KJoslyn: see note below in `applyColors()`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
```diff
@@ -0,0 +1,53 @@
...
+        } else {
+            applyColors(secondaryColor, primaryColor)
+        }
+    }
+
+    private fun applyColors(bgColor: Color, contentFillColor: Color) {
+        style {
+            backgroundColor += bgColor
+            if(hasBorder) {
+                borderColor += box(primaryColor)
```
KJoslyn: I believe the borderColor should always be set to the primary color if there is a border, because setting the border color to the secondary color (presumably white) causes issues with the background. Also, the border color cannot simply be set once when the button is instantiated, because each time `style` is set here, previous styles get wiped out. Thus, border color must be set here in the parent class, every time `applyColors` is called.

aunger (approved):

jsarabia (changes requested):
I'll let Nathan take a look and make suggestions before I make any of my own, but I think that there's more expected ways to implement/use this that is consistent with other UI Controls, such as Button itself.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
```diff
@@ -0,0 +1,60 @@
...
+import de.jensd.fx.glyphs.materialicons.MaterialIconView
+import javafx.beans.property.SimpleBooleanProperty
+import javafx.scene.paint.Color
+import org.wycliffeassociates.otter.jvm.app.theme.AppTheme
+import tornadofx.*
+
+class HighlightableButton : JFXButton() {
+
+    var primaryColor: Color = AppTheme.colors.black
+    var secondaryColor: Color = AppTheme.colors.white
```
jsarabia: Don't make this dependent on AppTheme; just do Color.white/Color.black

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
natibekele: what do you think about giving these function different names?  so that they're both not `applyColors()`

KJoslyn: Sure, I'd probably name the private one something like `doApplyColors()` then

jsarabia: what is the += operator doing exactly? are you adding colors? are you appending colors to a list? is there a reason it's not just =?

natibekele: The reason we have to use the `+=` is because the javafx css is expecting a type of  `MutliValue<Paint>` for the backgroundColor. If you jsut use a `=` you get a Type Mismatch

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/ResourceCard.kt
```diff
@@ -0,0 +1,77 @@
...
+    rc.init()
+    return rc
+}
+
+private fun StatusIndicator.initForResourceCard() {
+    prefHeight = 5.0
+    prefWidth = 75.0
+    primaryFill = AppTheme.colors.appOrange
+    accentFill = AppTheme.colors.lightBackground
+    trackFill = AppTheme.colors.defaultBackground
```
natibekele: I think is the same as what joe said in another comment. Don't make the ResouceCard control dependent on the colors coming from the AppTheme. You could set the default values here for the fills with `Color.(whateverColor)` and then expose some ColorProperties that you could expose on the ResourceCard itself. and then bind those to the primaryFill,accentFill and trackFill of the statusIndicator. That way if anyone is using the ResourceCard and wants to change the colors of the statusinidcators they can

natibekele: maybe `progress properties` too ?

KJoslyn: I'm not sure how we're going to implement progress yet, so maybe a progress property should wait for later? 

I think a color property on ResourceCard for primaryColor/fill is a good idea. I'm not sure about creating properties for accentFill and trackFill yet though since a feature to change those colors on the resourcecard hasn't been established. Thoughts?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/highlightablebutton/HighlightableButton.kt
```diff
@@ -3,45 +3,42 @@ package org.wycliffeassociates.otter.jvm.app.widgets.highlightablebutton
...
 import javafx.beans.property.SimpleBooleanProperty
+import javafx.beans.property.SimpleObjectProperty
 import javafx.scene.paint.Color
 import tornadofx.*
 
 class HighlightableButton : JFXButton() {
 
-    var primaryColor: Color = Color.BLACK
+    var isHighlightedProperty: SimpleBooleanProperty = SimpleBooleanProperty(false)
+    var highlightColorProperty = SimpleObjectProperty<Color>(Color.BLACK)
```
KJoslyn, line 13: Renamed this to `highlightColorProperty` so it wouldn't conflict with `primaryColorProperty` in `ResourceCard`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/ResourceCard.kt
```diff
@@ -16,7 +18,9 @@ import tornadofx.FX.Companion.messages
 
 class ResourceCard(private val resource: Resource) : HBox() {
 
-    val isCurrentResource = SimpleBooleanProperty(false)
+    val isCurrentResourceProperty = SimpleBooleanProperty(false)
+    var primaryColorProperty = SimpleObjectProperty<Color>(Color.ORANGE)
+    var primaryColor: Color by primaryColorProperty
```
KJoslyn: NEW 4/5/19: To test this property, go to `ResourceGroupCard.kt` and change line 15 to
```
resourcecard(it) {
     primaryColor = Color.(X)
}
```

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/ResourceCard.kt
natibekele: Yes I actually I think you have a fair point. We don't have feature to change those colors yet, I was just thinking that a situation could arise where we might need a different color, but since that isn't specced out yet we can add that feature to it in the future if needed

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/ResourceCard.kt
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        region {
+            hgrow = Priority.ALWAYS
+        }
+
+        add(
+            highlightablebutton {
+                highlightColorProperty.bind(primaryColorProperty)
+                secondaryColor = AppTheme.colors.white
```
natibekele: same thing here. try to not have app dependencies in any of your `widgets/controls`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/ResourceGroupCardStyles.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    companion object {
+        val resourceGroupCard by cssclass()
+    }
+    init {
+        resourceGroupCard {
+            spacing = 10.px // VBox spacing
+            padding = box(15.px)
+            backgroundColor += AppTheme.colors.white
+            effect = DropShadow(2.0, 2.0, 4.0, AppTheme.colors.lightBackground)
```
natibekele: last couple. backgroundColor and effect have AppDependencies

KJoslyn: Aw geez sorry I shouldn't have assumed they were all gone!

jsarabia (approved):

natibekele (approved):

jsarabia (approved):

PR #441, opened 2019-04-05
WorkbookRepository

aunger:
Implement WorkbookRepository to build new data model objects.
Co-requires https://github.com/WycliffeAssociates/otter-common/pull/97


src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
```diff
@@ -0,0 +1,213 @@
...
+    private val collectionRepository: CollectionRepository,
+    private val contentRepository: ContentRepository,
+    private val resourceRepository: ResourceRepository,
+    private val takeRepository: TakeRepository
+) : IWorkbookRepository {
+    /** Disposers for Relays in the current workbook. */
+    private val connections = CompositeDisposable()
+
+    /** Map to recover model.Take objects from workbook.Take objects. */
+    private val takeMap = WeakHashMap<Take, org.wycliffeassociates.otter.common.data.model.Take>()
```
jsarabia: I suppose this works because the only takes added to the relay are what would be inserted? So even though null ids typically mean the take is new and can be inserted into the database, it's fine if they're lost through having to re-map a lost weakhashmap Take because we no longer need/care about the ids of existing Takes; updates to specific fields happen through their respective relays, and inserting new takes happens when a new Take is pushed into the relay (regardless of the contents of the id field)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
```diff
@@ -0,0 +1,213 @@
...
+            .subscribe { updatedSelection ->
+                content.selectedTake = updatedSelection?.let { modelTake(it) }
+                contentRepository.update(content)
+            }
+
+        connections += subscription
+        return relay
+    }
+
+    private fun constructTakes(content: Content): ReplayRelay<Take> {
```
jsarabia: This would also have to handle subscriptions to the deletedTimestamp field- both on the takes coming from the database but also on takes being inserted into the database

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
```diff
@@ -0,0 +1,213 @@
...
+
+        // Push Takes from DB into relay
+        takeRepository
+            .getByContent(content)
+            .flattenAsObservable { list -> list.sortedBy { it.number } }
+            .map { workbookTake(it) }
+            .subscribe(relay)
+
+        // Respond to new Takes pushed into relay by writing them to DB
+        val subscription = relay.subscribe { newTake ->
```
jsarabia: this could get executed as takes are coming out from the database I believe?

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
aunger: I scoped this to a single function so its purpose is more clear.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
aunger: Yes.  I think that's fixed now.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
```diff
@@ -0,0 +1,215 @@
...
+            // Insert the new take into the DB.
+            .subscribe { (_, modelTake) ->
+                takeRepository.insertForContent(modelTake, content)
+            }
+
+        /** The initial selected take, from the DB. */
+        val initialSelectedTake = content.selectedTake?.let { workbookTake(it) }
+
+        /** Relay to send selected-take updates out to consumers, but also receive updates from UI. */
+        val selectedTakeRelay = BehaviorRelay.createDefault(initialSelectedTake)
```
aunger: I just discovered RxJava2 streams break on null elements.  So our nullable observables will need to be wrapped in Optionals.  Grr.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceContainerRepository.kt
```diff
@@ -64,36 +68,37 @@ class ResourceContainerRepository(
...
+        creator: String,
+        dsl: DSLContext
+    ): List<Int> {
         val relatedIds = mutableListOf<Int>()
         relations.forEach { relation ->
             val parts = relation.split('/')
             // NOTE: We look for derivedFromFk=null since we are looking for the original resource container
             resourceMetadataDao.fetchLatestVersion(parts[0], parts[1], creator, null, dsl)
-                    ?.let { relatedDublinCore ->
-                        // TODO: Only add link if it doesn't exist already
```
aunger, line 80: Todo is handled via new ON CONFLICT IGNORE on DB constraint

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/database/daos/ContentTypeDao.kt
```diff
@@ -0,0 +1,58 @@
+package org.wycliffeassociates.otter.jvm.persistence.database.daos
```
aunger, line 1: The first time a `get` or `fetch` is called on this Dao, it loads the DB indices into memory (and creates them upon first run) and exposes them as an extension property on the `common.data.model.ContentType` enum.

src/main/resources/sql/CreateAppDb.sql
```diff
@@ -96,6 +96,12 @@ CREATE TABLE IF NOT EXISTS resource_link (
         CHECK ((collection_fk is NULL) or (content_fk is NULL))
 );
 
+CREATE TABLE IF NOT EXISTS subtree_has_resource (
+  collection_fk       INTEGER NOT NULL REFERENCES collection_entity (id) ON DELETE CASCADE,
+  dublin_core_fk      INTEGER NOT NULL REFERENCES dublin_core_entity (id) ON DELETE CASCADE,
+  PRIMARY KEY (collection_fk, dublin_core_fk) ON CONFLICT IGNORE
+);
+
```
aunger, line 113: This is a denormalized table that lets us quickly look up which helps resources are available for a collection or any of its children.

src/main/resources/sql/CreateAppDb.sql
```diff
@@ -95,6 +95,8 @@ CREATE TABLE IF NOT EXISTS resource_link (
     CONSTRAINT prevent_both_not_null
         CHECK ((collection_fk is NULL) or (content_fk is NULL))
 );
+CREATE INDEX IF NOT EXISTS idx_resource_link_collection ON resource_link (collection_fk);
+CREATE INDEX IF NOT EXISTS idx_resource_link_content ON resource_link (content_fk);
```
aunger, line 106: I'm always embarrassingly slow to realize a tiny DB tweak is worth many hours of code optimizations.  These lines dropped import time (with the new subtreeHasResources lookup) from many minutes back down to seconds.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
```diff
@@ -160,7 +160,7 @@ class WorkbookRepository(
             .filter { localDate -> localDate.value != null }
             .filter { take == selectedTakeRelay.value?.value }
             .map { null }
-            .subscribe { selectedTakeRelay }
```
aunger: Static type-checking wasn't good enough here, but I'm SO GLAD we're not in a dynamic language.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
aunger: Resolved in e898910

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
```diff
@@ -0,0 +1,216 @@
...
+            .map { workbookTake(it) to it }
+
+        /** Relay to send Takes out to consumers, but also receive new Takes from UI. */
+        val takesRelay = ReplayRelay.create<Take>()
+        takesFromDb
+            // Record the mapping between data types.
+            .doOnNext { (wbTake, modelTake) -> takeMap[wbTake] = modelTake }
+            // Feed the initial list to takesRelay
+            .map { (wbTake, _) -> wbTake }
+            .subscribe(takesRelay)
```
KJoslyn, line 233: Trying to make sure I understand what's going on- am I right to say that the TakesRelay will receive a take from the DB (converted to a wbTake), and emit that take so that the UI can respond and display it? The TakeRelay will not do anything else with takes received from the DB since they will already be in the takeMap.

src/main/resources/sql/CreateAppDb.sql
KJoslyn: I just did a quick search on denormalized tables. By this you mean that the table essentially duplicates information that we could find in the database, but the intention of the table is to avoid digging deep into multiple tables to see if there are resources for a collection. Correct?

KJoslyn: Also why "ON CONFLICT IGNORE"? 
UPDATE: Never mind, I thought "ignore" meant to insert anyway- it means the opposite.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
aunger: @KJoslyn Yes, but not just "a take". It will be zero, one, or many takes. But as you said, they're only loaded from the DB once. Each relay might get updates from the UI layer, but never again from the DB.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ContentRepository.kt
```diff
@@ -32,6 +35,17 @@ class ContentRepository(
                 .subscribeOn(Schedulers.io())
     }
 
+    override fun getCollectionMetaContent(collection: Collection): Single<Content> {
+        return Single
+            .fromCallable {
+                contentDao
+                    .fetchByCollectionIdAndStart(collection.id, 1, listOf(ContentDao.Labels.CHAPTER))
```
KJoslyn, line 42: Why is the start parameter 1 here? Perhaps I don't remember what this meta content is.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
```diff
@@ -0,0 +1,259 @@
...
+        connections += subscription
+        return relay
+    }
+
+    private fun deselectUponDelete(take: WorkbookTake, selectedTakeRelay: BehaviorRelay<TakeHolder>) {
+        val subscription = take.deletedTimestamp
+            .filter { localDate -> localDate.value != null }
+            .filter { take == selectedTakeRelay.value?.value }
+            .map { TakeHolder(null) }
+            .subscribe(selectedTakeRelay)
```
KJoslyn, line 174: IMO, the concept behind `deselectUponDelete` is brilliant, but difficult for me personally to understand without comments. Though I think it is difficult simply because I dislike the way that `subscribe(consumer)` LOOKS syntactically. The way I think of it is that selectedTakeRelay is subscribing TO deletedTimestamp, but this makes it look like deletedTimestamp is subscribing to selectedTakeRelay. I would have named the `subscribe` method something more unambiguous, but perhaps it just takes a little getting used to!

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ContentRepository.kt
jsarabia: The parameter is for the start, but this is the metacontent so the start would always be 1

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/WorkbookRepository.kt
aunger: I agree.  `a.subscribe(b)` should mean `a` wants to be notified of `b`'s events.  That's not how Rx uses it though.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ContentRepository.kt
aunger: This line will also be replaced to select `ContentType.META` (in another PR).

jsarabia (approved):

PR #442, opened 2019-04-08
consolidate-grids

natibekele:
draft pr for merging/consolidating `collectiongrid` and `contentgrid` into one
`cardgrid`


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/cardgrid/CardData.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.wycliffeassociates.otter.common.data.model.Collection
+import org.wycliffeassociates.otter.common.data.model.Content
+
+data class CardData(
+    val item: String,
+    val dataType: String,
+    val bodyText : String,
+    val sort: Int,
+    val contentSource: Content? =null,
+    val collectionSource: Collection? =null
```
jsarabia, line 12: These cards don't need to know about the underlying content they represent.

What I mean by that is, they provide the appropriate places to set content to, and whatever configures it (binds content to the UI) can know about the actual domain stuff.

Like, The card cares about major text, minor text, button text/icon, graphic, whatever. It doesn't need to know that the major text is ACTUALLY a book title, it doesn't care, it just cares that you told it this is the text you want in the top slot.

jsarabia (changes requested):
I feel like you're blurring the line too much between app functionality and views.

The way I see it, you have:
BookCard, (can't think of a good name right now, it is one card for books, chapters, and chunks)
BookCardFragment (takes care of adding the datagrid or w/e, adding the styling, handles whatever is necessary for responsiveness)

BookNavigationViewModel here binds your Books, Chapters, Chunks to whatever APIs you need to set the content in the cards and datagrid's adapter. This is why the cards should be agnostic of the content, because the datagrid will create the necessary cells (for us, cards) based on whatever elements and however many elements are in the list. This is why we didn't need a separate fragment or view- the view is the same, we just swap the underlying data that is bound to the view.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/mainscreen/view/MainScreenView.kt
```diff
@@ -112,12 +112,14 @@ class MainScreenView : View() {
         //navigate back to verse selection from viewing takes
         if (viewModel.selectedContentProperty.value != null) {
             viewModel.selectedContentProperty.value = null
+            activeFragment.navigateBack()
         }
         //from verse selection, navigate back to chapter selection
         else if (viewModel.selectedCollectionProperty.value != null) {
```
aunger, line 118: Why shouldn't navigateBack() be called in this case?

natibekele: because when we navigate from the chapter cards to the verse/chunk cards we don't actually dock a new fragment. Collection grid and content grid were consolidated into one view model, and then we have one `card data` data type which was determining the data that is displayed in the datagrid. so when a user navigated from the chapters to verses the underlying data was changing but the fragment stayed the same

jsarabia (approved):

PR #447, opened 2019-04-18
Workbook Header

KJoslyn:
Header for the upcoming workbook fragment. Includes title, progress bar, and filter checkbox.

The filter checkbox doesn't do anything yet, but you can set a Predicate<T> to be used by the viewmodel (later) to filter the list.

To test, please check out [kj-workbook-header-testapp](https://github.com/WycliffeAssociates/otter-jvm/tree/kj-workbook-header-testapp) and run `main` in **ResourcesCardApp.kt** (this reuses the same test application that I used for the resource cards.)

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/workbookheader/WorkbookHeader.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/workbookheader/WorkbookHeader.kt
@@ -1,0 +1,83 @@
+package org.wycliffeassociates.otter.jvm.app.widgets.workbookheader
+
+import com.jfoenix.controls.JFXCheckBox
+import javafx.beans.property.SimpleDoubleProperty
+import javafx.beans.property.SimpleObjectProperty
+import javafx.beans.property.SimpleStringProperty
+import javafx.scene.layout.Priority
+import javafx.scene.layout.VBox
+import javafx.scene.paint.Color
+import org.wycliffeassociates.otter.jvm.statusindicator.control.statusindicator
+import tornadofx.*
+import java.util.function.Predicate
+
+data class FilterOption<T>(val text: String, val predicate: Predicate<T>)
+
+class WorkbookHeader<T> : VBox() {
+
+    val labelTextProperty = SimpleStringProperty()
+    var labelText by labelTextProperty
+
+    val filterOptionProperty = SimpleObjectProperty<FilterOption<T>?>()
+    var filterOption by filterOptionProperty
+
+    val progressBarTrackFillProperty = SimpleObjectProperty<Color>(Color.ORANGE)
+    var progressBarTrackFill by progressBarTrackFillProperty
+
+    val workbookProgressProperty = SimpleDoubleProperty(0.5)
+    var workbookProgress by workbookProgressProperty
+
+    private val filterOptionTextProperty = SimpleStringProperty()
+
+    init {
+        importStylesheet<WorkbookHeaderStyles>()
+
+        filterOptionProperty.onChange {
+            filterOptionTextProperty.set(it?.text)
+        }
+
+        addClass(WorkbookHeaderStyles.workbookHeader)
+        spacing = 10.0
+        hbox {
+            label(labelTextProperty) {
+                managedProperty().bind(!labelTextProperty.isEmpty)
+            }
+            region {
+                hgrow = Priority.ALWAYS
+            }
+            add(
+                JFXCheckBox().apply {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/workbookheader/WorkbookHeader.kt
```diff
@@ -0,0 +1,83 @@
...
+                JFXCheckBox().apply {
+                    isDisableVisualFocus = true
+                    textProperty().bind(filterOptionTextProperty)
+                    managedProperty().bind(filterOptionProperty.isNotNull)
+                    visibleProperty().bind(filterOptionProperty.isNotNull)
+                    action {
+                        if (isSelected) {
+                            // TODO: Tell view model to filter list
+                        } else {
+                            // TODO: Tell view model to remove filter from list
```
aunger: Why TODO? Shouldn't WorkbookHeader just set a boolean property on itself that can be bound to whatever does the filtering? I probably just don't understand something about the architecture here; let me know if I'm off base.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/workbookheader/WorkbookHeader.kt
```diff
@@ -0,0 +1,83 @@
...
+import javafx.beans.property.SimpleObjectProperty
+import javafx.beans.property.SimpleStringProperty
+import javafx.scene.layout.Priority
+import javafx.scene.layout.VBox
+import javafx.scene.paint.Color
+import org.wycliffeassociates.otter.jvm.statusindicator.control.statusindicator
+import tornadofx.*
+import java.util.function.Predicate
+
+data class FilterOption<T>(val text: String, val predicate: Predicate<T>)
```
aunger: It feels odd to me to store the filtering rules (predicate) in the header widget.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/workbookheader/WorkbookHeader.kt
KJoslyn: Well I am still making an assumption about the view model, which is that the workbook header and the workbook page (i.e. the part that holds the resources list or the chunks list) will share a view model. In that case, the view model will hold the list of resources or chunks. I think the view model would have a `isFilteredProperty` that should be bound to the `selectedProperty` of the checkbox so that the view model can filter the list. So all of this to say, you're probably right :)

natibekele: so why not just mock that isFIlteredProperty in the `Workbook Header`  class and use that selectedProperty so that is't closer to what it will be once the viewModel is actually made

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/workbookheader/WorkbookHeader.kt
```diff
@@ -0,0 +1,83 @@
...
+            statusindicator {
+                hgrow = Priority.ALWAYS
+                primaryFillProperty().bind(progressBarTrackFillProperty)
+                accentFillProperty.bind(progressBarTrackFillProperty)
+                trackFill = Color.LIGHTGRAY
+                prefHeight = 15.0
+                barHeight = 18.0
+                indicatorRadius = 10.0
+                progressProperty().bind(workbookProgressProperty)
+            }
```
natibekele, line 68: also, i have a pr up for the statusindicator with updated properties.  would you want to wait for that to merged in so that you can use those properties to finalize the statusindicator instance here

KJoslyn: yeah sounds good

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/workbookheader/WorkbookHeader.kt
jsarabia: Agreed, the widget should allow for:
 - customizing the filter checkbox text
 - exposing a property to observe the filter state (on or off)
 - changing filter state

Views are basically never the place to house real work being done

jsarabia (changes requested):
Remove filter rules from the widget, set properties for filtering

jsarabia (approved):

PR #453, opened 2019-05-20
Loading resources from workbook

KJoslyn:
To test, please check [kj-resources-list-testapp](https://github.com/WycliffeAssociates/otter-jvm/tree/kj-resources-list-testapp) and run `main` in `ResourceCardApp.kt`. This branch is a separate entry point to load a single chapter using the workbook repository.

**UPDATE 6/3/19: This PR once again depends on [#100](https://github.com/WycliffeAssociates/otter-common/pull/100) in common.


build.gradle
```diff
@@ -129,6 +129,9 @@ dependencies {
     implementation 'org.wycliffeassociates:8woc2018-common'
     implementation 'org.wycliffeassociates:kotlin-resource-container'
     implementation 'com.github.WycliffeAssociates:jdenticon-kotlin:-SNAPSHOT'
+
+    //Atlassian commonmark (for rendering markdown)
+    implementation 'com.atlassian.commonmark:commonmark:0.12.1'
```
KJoslyn, line 134: Uses the BSD 2-clause license. Should be fine, but I'm not sure where/how we're keeping track of licenses.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceRepository.kt
```diff
@@ -87,14 +89,41 @@ class ResourceRepository(private val database: AppDatabase) : IResourceRepositor
             .map(this::buildResourceInfo)
     }
 
-    override fun getResources(collection: Collection, resourceInfo: ResourceInfo): Observable<Content> {
+    /**
+     * Returns all resources for which the resource content's COLLECTION_FK field references this collection.
+     * This will return resources about the chapter as well as all resources about the chapter's chunks.
+     */
+    override fun getResourcesForCollectionAndChildren(
```
KJoslyn: see note in the corresponding common PR about these methods [#100](https://github.com/WycliffeAssociates/otter-common/pull/100)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceRepository.kt
```diff
@@ -87,14 +89,41 @@ class ResourceRepository(private val database: AppDatabase) : IResourceRepositor
...
     }
 
     override fun getResources(content: Content, resourceInfo: ResourceInfo): Observable<Content> {
         return getResources({ table -> table.ID.eq(content.id) }, resourceInfo)
     }
 
+    /**
+     * Returns collection-specific resources
+     */
+    override fun getResources(collection: Collection, resourceInfo: ResourceInfo): Observable<Content> {
```
KJoslyn: see note in the corresponding common PR about these methods #100

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/view/ResourceListFragment.kt
```diff
@@ -0,0 +1,31 @@
...
+        importStylesheet<ResourceListStyles>()
+    }
+    override val root = vbox {
+
+        addClass(MainScreenStyles.main)
+
+        add(
+            workbookheader {
+                labelText = viewModel.chapter.title + " Resources"
+                filterText = "Hide Completed"
```
aunger: i18n

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+        get() = activeResourceSlugProperty.value
+
+    var resourceGroups: ObservableList<ResourceGroupCardItem> = FXCollections.observableArrayList()
+
+    fun loadResourceGroups() {
+        chapter.chunks.map { chunk ->
+            resourceGroupCardItem(chunk, resourceSlug) { navigateToTakesPage(it) }
+        }.startWith(
+            resourceGroupCardItem(chapter, resourceSlug) { navigateToTakesPage(it) }
+        ).buffer(2).subscribe { // Buffering by 2 prevents the list UI from jumping while groups are loading
```
aunger: Is 2 the magic number because two resourcegroupcarditems can fit on the screen at once? Should this be a larger number if the items are small?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
```diff
@@ -1,5 +1,33 @@
...
+    val titleProgressProperty: DoubleProperty = resource.titleAudio.progressProperty()
+    val bodyProgressProperty: DoubleProperty? = resource.bodyAudio?.progressProperty()
+    val hasBodyAudio: Boolean = resource.bodyAudio != null
+
+    private fun AssociatedAudio.progressProperty(): DoubleProperty {
+        val progressProperty = SimpleDoubleProperty(0.0)
+        this.selected.subscribe {
+            progressProperty.set( if (it.value != null) 1.0 else 0.0)
+        }
+        return progressProperty
```
aunger, line 33: I'm having a hard time thinking this through, but I'm uneasy about whether this will create a memory leak. You're giving the `ResourceCardItem` a `DoubleProperty` which is going to be held by the `resource.titleAudio.selected` Observable's subscription.  I believe this will persist until the whole workbook is unloaded (roughly forever).

I _think_ you just need to kill this subscription (and the one for body) when the UI item goes away. I could be wrong about this whole thing though.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
```diff
@@ -1,5 +1,33 @@
...
+
+    companion object {
+        val parser: Parser = Parser.builder().build()
+        val renderer: TextContentRenderer = TextContentRenderer.builder().build()
+    }
+
+    private fun getTitleTextContent(): String {
+        val document = parser.parse(resource.title.text)
+        return renderer.render(document)
+    }
```
aunger, line 44: It's not really a getter, so I'd rename this to something like "renderTitleAsHtml".

aunger: It's not needed yet, but the second time you need to transform MD to HTML, you should probably pull this commonmark stuff out into an dagger-injectable MarkdownToHtml singleton.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceGroupCardItem.kt
```diff
@@ -1,8 +1,40 @@
...
+private fun findResourceGroup(element: BookElement, slug: String): ResourceGroup? {
+    return element.resources.firstOrNull {
+        it.info.slug == slug
+    }
+}
+
+private fun getGroupTitle(element: BookElement): String {
+    return when (element) {
+        is Chapter -> "Chapter " + element.title
+        is Chunk -> "Chunk " + element.title
```
aunger: i18n

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/view/ResourceGroupCard.kt
```diff
@@ -10,10 +11,15 @@ class ResourceGroupCard(group: ResourceGroupCardItem) : VBox() {
 
         addClass(ResourceGroupCardStyles.resourceGroupCard)
         label(group.title)
-        group.resources.subscribe {
-            add(
-                resourcecard(it)
-            )
+
+        group.resources.buffer(10).subscribe { items ->
```
aunger: Please make this a named constant. (Maybe `RENDER_BATCH_SIZE`, or `ITEMS_VISIBLE_IN_ONE_SCREEN`, if that's indeed what it is.)

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceRepository.kt
aunger: I don't think we actually need `getResourcesForCollectionAndChildren`.  I think your new `getResources` behavior is all that's needed now.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceRepository.kt
```diff
@@ -87,14 +89,41 @@ class ResourceRepository(private val database: AppDatabase) : IResourceRepositor
...
+            .selectDistinct(help.asterisk())
+            .from(RESOURCE_LINK)
+            .join(COLLECTION_ENTITY).on(COLLECTION_ENTITY.ID.eq(RESOURCE_LINK.COLLECTION_FK))
+            .join(help).on(RESOURCE_LINK.RESOURCE_CONTENT_FK.eq(help.ID))
+            .where(RESOURCE_LINK.DUBLIN_CORE_FK.eq(metadata.id))
+            .and(COLLECTION_ENTITY.ID.eq(collection.id))
+
+        return getResources(help, selectStatement)
+    }
+
```
aunger, line 108: I mean, I know I fully deserve this, but this stuff is _hard to read_.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    val activeResourceSlugProperty = SimpleStringProperty()
+    val resourceSlug: String
+        get() = activeResourceSlugProperty.value
+
+    var resourceGroups: ObservableList<ResourceGroupCardItem> = FXCollections.observableArrayList()
+
+    fun loadResourceGroups() {
+        chapter.chunks.map { chunk ->
+            resourceGroupCardItem(chunk, resourceSlug) { navigateToTakesPage(it) }
```
aunger: RxJava2 Observables will crash on null elements. `resourceGroupCardItem` can return nulls, so guard against it. Maybe you can think of a better way, but I took a crack at it since it seemed fun:
```
fun <T, R> Observable<T>.mapNotNull(f: (T) -> R?): Observable<R> =
    concatMapMaybe { Maybe.fromCallable { f(it) } }

fun loadResourceGroups() {
    chapter
        .children
        .startWith(chapter)
        .mapNotNull { resourceGroupCardItem(it, resourceSlug, onSelect = this::navigateToTakesPage) }
        .buffer(2) // Buffering by 2 prevents the list UI from jumping while groups are loading
        .subscribe {
            Platform.runLater {
                resourceGroups.addAll(it)
            }
        }
}
```

aunger (changes requested):

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
aunger: Maybe better than Maybe :
```
fun <T, R> Observable<T>.mapNotNull(f: (T) -> R?): Observable<R> =
    concatMapIterable { listOfNotNull(f(it)) }
```

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: Certainly more resourcegroupcards can fit on the screen if they are small- upwards of 7 in fact. I just did a little experiment adding very small resource group cards with dummy data, but I did not observe any UI jumping whether I buffered the cards or not. When I experimented loading several different chapters with real data, buffering by 2 works every time even when more than 2 group cards fit on the screen. It is possible that later we could find that it doesn't work in every case, but I'm not sure how to uncover that right now.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
KJoslyn: Good points. I'm actually just rendering the title as plain text though and saving html rendering for the body when it appears on the takes page.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/view/ResourceGroupCard.kt
KJoslyn: Sure. I created this as a render batch size since I thought it was too chaotic to see each resource get added individually.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: The second one requires that `R` extends `Any` for the compiler to accept, but I might still prefer that one.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceRepository.kt
KJoslyn: Should be better now that I removed `getResourcesForCollectionAndChildren`.

src/main/kotlin/org/wycliffeassociates/otter/jvm/persistence/repositories/ResourceRepository.kt
KJoslyn: The only reason I can think that we might want it is if we want to load all of the resources at once for a chapter to calculate progress. But we haven't discussed how we'll do that yet, and this function wouldn't be totally necessary anyway. So I'll remove it.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/view/ResourceListFragment.kt
KJoslyn: Done.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceGroupCardItem.kt
KJoslyn: done.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
KJoslyn: Good point. To be safe, I created a `ResourceGroupCardItemList` class that listens to changes in the observable list of ResourceGroups and kills the subscriptions when a group is removed.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/view/ResourceListFragment.kt
```diff
@@ -0,0 +1,31 @@
...
+        importStylesheet<MainScreenStyles>()
+        importStylesheet<ResourceListStyles>()
+    }
+    override val root = vbox {
+
+        addClass(MainScreenStyles.main)
+
+        add(
+            workbookheader {
+                labelText = "${viewModel.chapter.title} ${messages["resources"]}"
```
aunger, line 23: i18n shouldn't be done word-by-word. This code will yield "capítulo 1 recursos" in Spanish, but it should be "recursos del capítulo 1".

A quick look at the resource file makes me think this hasn't come up before. We probably want to use [java.text.MessageFormat](https://docs.oracle.com/javase/8/docs/api/java/text/MessageFormat.html). I think you could do it now if you want, but it's probably better to break this off as its own task.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
aunger: Will items be removed from the `ResourceGroupCardItemList`? If the list is just thrown away, does it notify its contents of their removal?  If so, this should work.  But the alternative might still be cleaner:

Would it be better to just do `ResourceCard.onUndock() { ResourceCardItem.clearDisposables() }`?  If `onDock` and `onUndock` might be called multiple times, you might also need `ResourceCard.onDock() { ResourceCardItem.subscribeToModel() }`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/view/ResourceGroupCard.kt
```diff
@@ -6,13 +6,14 @@ import org.wycliffeassociates.otter.jvm.app.widgets.resourcecard.model.ResourceG
 import tornadofx.*
 
 class ResourceGroupCard(group: ResourceGroupCardItem) : VBox() {
+    private val RENDER_BATCH_SIZE = 10
```
aunger: Can make this a const if it's at the file level or in the companion object, I think.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
aunger: Cool.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
KJoslyn: Unfortunately it appears that only UIComponents (Fragments and Views) have `onDock` and `onUndock` methods and after spending some time trying to figure out if there is an equivalent set of methods for Node objects, I haven't found a solution that works.

I don't mean to sound hand-wavy about this, but would the `ResourceGroupCardItemList` ever be thrown away? I believe if it is a non-nullable val inside a ViewModel (which is a singleton and only initialized once), it will not.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/view/ResourceListFragment.kt
KJoslyn: I agree, this should probably be it's own task.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/view/ResourceGroupCard.kt
KJoslyn: I think I like it in a companion object since it is specific to the resource group card.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
aunger: Cool, I didn't realize `ResourcesViewModel` is a singleton.  That means it won't keep leaking more objects.  

Can I ask that you add one more thing to be sure the last set of subscriptions go away if the user exits this part of the app?
```
@Suppress("ProtectedInFinal", "Unused") protected fun finalize() { clearDisposables() }
```
(See [Stack Overflow](https://stackoverflow.com/questions/43784161/how-to-implement-finalize-in-kotlin).)

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/view/ResourceListFragment.kt
aunger: OK, spawned https://github.com/WycliffeAssociates/otter-jvm/issues/454 / [Zenhub](https://app.zenhub.com/workspaces/otter-jvm-5b3f8f4e9e645c042e87275c/issues/wycliffeassociates/otter-jvm/454)

aunger (changes requested):

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
aunger: BTW, I don't _think_ these are real `@Singleton`s, so when the ViewModel isn't used anymore, Dagger should release it and all this should be finalized. http://frogermcs.github.io/dependency-injection-with-dagger-2-custom-scopes/

KJoslyn: Cool, thanks for the links! Made that change too.

aunger (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
```diff
@@ -0,0 +1,46 @@
...
+    val chapter: Chapter
+        get() = activeChapterProperty.value
+
+    val activeResourceSlugProperty = SimpleStringProperty()
+    val resourceSlug: String
+        get() = activeResourceSlugProperty.value
+
+    val resourceGroups: ResourceGroupCardItemList = ResourceGroupCardItemList(mutableListOf())
+
+    fun <T, R:Any> Observable<T>.mapNotNull(f: (T) -> R?): Observable<R> =
```
jsarabia: this should probably just be moved out to its own file in a utils package

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
KJoslyn: @sarabiaj I would like to talk to you about dagger-injectable components like this, as opposed to an object in common, like ParseMd.kt.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
```diff
@@ -1,5 +1,45 @@
...
+data class ResourceCardItem(val resource: Resource, val onSelect: () -> Unit) {
+    val title: String = renderTitleAsPlainText()
+    private val disposables = CompositeDisposable()
+    val titleProgressProperty: DoubleProperty = resource.titleAudio.progressProperty()
+    val bodyProgressProperty: DoubleProperty? = resource.bodyAudio?.progressProperty()
+    val hasBodyAudio: Boolean = resource.bodyAudio != null
+
+    @Suppress("ProtectedInFinal", "Unused")
+    protected fun finalize() {
+        clearDisposables()
```
jsarabia, line 20: could you try putting a print statement here and seeing if this actually executes? I'm kinda concerned about the subscribe reference not allowing this to be marked for GC?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: Where would you like to see this go? Should a utils package be made in common or does it need to be under another package such as observable > utils?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
KJoslyn: This does in fact execute when the `ResourceGroupCardItem`s are removed from the observable list, even when the `ResourceCardItem`s have subscriptions to `AssociatedAudio.selected`. Since this is the case (and I should've realized this sooner), I suppose I don't need any of the additional logic that I added to manually clear the subscriptions. However, this was only tested when the `ResourceGroupCardItem` list is cleared programmatically- I don't know how to test it if the list is not cleared, i.e. how do I just drop the reference to the list or the ViewModel? 

Edit: Although there may be a way to test this by undocking the component that is dependent on the ViewModel (maybe), I feel fairly confident that finalize() would run, since clearing the list of `ResourceGroupCardItem`s (and effectively dropping the references to the `ResourceGroupCardItem`s) caused finalize() to be called on the associated `ResourceCardItem`s when they were GC'd. I can still attempt to test if this is desired.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
jsarabia: put it in common.utils

maybe like RxUtils.kt

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
```diff
@@ -0,0 +1,46 @@
...
+        concatMapIterable { listOfNotNull(f(it)) }
+
+    fun loadResourceGroups() {
+        chapter
+            .children
+            .startWith(chapter)
+            .mapNotNull { resourceGroupCardItem(it, resourceSlug, onSelect = this::navigateToTakesPage) }
+            .buffer(2) // Buffering by 2 prevents the list UI from jumping while groups are loading
```
jsarabia, line 32: what happens if there's an odd number of resource groups?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
jsarabia: I'd prefer to keep dagger to injection in the framework classes (Android Activities/Fragments, TornadoFX Views/ Fragments, maybe viewmodels)

for a component like this, I'd rather you just pass the dependency into the constructor rather than make dagger do it for you

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: When the source observable completes, the observable will emit whatever is in the current buffer, so it won't matter if there's an odd number of resource groups.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
KJoslyn: Sure. In this case, it seems like this can be solved without passing a dependency into the constructor. I think I can just create a MarkdownToHtml object like ParseMd that basically just holds static functions to do what I need.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/model/ResourceCardItem.kt
jsarabia: I just read through the resolved discussion and realized that what makes me uncomfortable here is that clearDisposables() needs to be tied to a lifecycle API, and as you both discussed, onDock/onUndock are good lifecycle places to attach this to. In the absence of one, you're attaching to the GC lifecycle whether consciously or unconsciously.

The CardItem here isn't exactly a framework object, and as such has no APIs to make use of- so this functionality would have to be exposed as a public method here. I'm going to look and see if ListView has some kind of lifecycle APIs to clean cells that may be more appropriate?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/view/ResourceListView.kt
```diff
@@ -0,0 +1,21 @@
...
+import javafx.collections.ObservableList
+import javafx.scene.control.ListView
+import javafx.scene.layout.Priority
+import org.wycliffeassociates.otter.jvm.app.widgets.resourcecard.model.ResourceGroupCardItem
+import org.wycliffeassociates.otter.jvm.app.widgets.resourcecard.styles.ResourceListStyles
+import tornadofx.*
+
+class ResourceListView(items: ObservableList<ResourceGroupCardItem>): ListView<ResourceGroupCardItem>(items) {
+    init {
+        vgrow = Priority.ALWAYS // This needs to be here
```
jsarabia: can this go in the stylesheet?

jsarabia (changes requested):
Please move the util function out and take a look at the style comment and you should be good to go after that

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/view/ResourceListView.kt
KJoslyn: No, vgrow is a property that cannot be set in the stylesheet.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: done

aunger: @KJoslyn If it's a util, I guess it should have kdoc:
```
/**
 * Apply the given function as with `map()`, but remove any resulting nulls from the
 * observable stream.
 * 
 * This is how `observable.map(f).filter { it != null }` should behave, but it keeps
 * nulls from ever appearing in an Observable, which would cause a crash in that
 * two-step map/filter example.
 */
```

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/widgets/resourcecard/view/ResourceListView.kt
jsarabia: Ahh, thanks!

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: Awesome, thanks! Just added that.

PR #458, opened 2019-06-14
Takes View Model- JVM

KJoslyn:
**What is included:**
- Takes tabpane is now populated with the selected resource's content and takes
- "New Take" button opens the preferred audio plugin and creates a new take for the selected resource
- Only one take can be played at a time. If a take is playing and another take is played, the first one will pause.

**What is not included:**
- Editing, deleting, selecting takes
- Properly formatted markdown (what you see is the raw markdown content)
- Navigation (you can navigate between the resources list and takes page by checking out the kj-takes-testapp branch, but this code is not in the PR.)

**How to test:**
1. Check out `kj-takes-testapp` branch (based on this PR's branch), and run `main` in ResourceTakesApp.kt. Also make sure to checkout the corresponding branch in common, `kj-takes-tabpane-newtake` (sister PR [#105](https://github.com/WycliffeAssociates/otter-common/pull/105) in common)
2. A chapter will be loaded by default, after which you can select "View Recordings" to go to the takes page for any resource. 
3. Try creating new takes
4. Play a take, and try playing a take while another take is playing
5. Switch between the "snippet" and "note" tabs
6. Return to the resource list page by clicking the "back" button in the header. (This header as well as the file menu are not in this PR, they are only in the kj-takes-testapp branch.)


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/viewmodel/TakesViewModel.kt
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+    }
+
+    fun onTabSelect(recordable: Recordable) {
+        activeRecordable = recordable
+    }
+
+    fun setRecordableListItems(items: List<Recordable>) {
+        if (!recordableList.containsAll(items))
```
KJoslyn: This makes use of the overriden equals() method on Recordable

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/workbook/viewmodel/WorkbookViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.wycliffeassociates.otter.common.data.workbook.Chunk
+import org.wycliffeassociates.otter.common.data.workbook.Workbook
+import tornadofx.*
+import java.io.File
+import java.lang.IllegalStateException
+
+class WorkbookViewModel: ViewModel() {
+    val activeWorkbookProperty = SimpleObjectProperty<Workbook?>()
+    val workbook: Workbook
+        get() = activeWorkbookProperty.value ?: throw IllegalStateException("Workbook is null")
```
KJoslyn, line 15: Rather than performing null (or ?.let) checks everywhere and dirtying up the code, I elected to make these getters throw errors if the value is null.
Since all of the getters in WorkbookViewModel follow this pattern, there should be no surprises as long as one knows this is the expected behavior.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/TakeManagementViewModel.kt
```diff
@@ -31,6 +32,9 @@ class TakeManagementViewModel : ViewModel() {
     private val takeRepository = injector.takeRepository
     private val pluginRepository = injector.pluginRepository
 
+    private val takesViewModel: TakesViewModel by inject()
+    private val workbookViewModel: WorkbookViewModel by inject()
```
KJoslyn, line 33: These are added so that we can call the new record() function (line 160.) However, TakeManagementViewModel and TakesViewModel will be merged soon, so the changes to this file will be obsolete when that happens.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesStyles.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    init {
+        main {
+            prefWidth = Screen.getPrimary().visualBounds.width.px - 600.0
+            prefHeight = Screen.getPrimary().visualBounds.height.px - 200.0
+            backgroundColor += AppTheme.colors.white
+        }
+
+        leftRegionContainer {
+            borderColor += box(Color.TRANSPARENT, AppTheme.colors.lightBackground, AppTheme.colors.lightBackground, Color.TRANSPARENT)
```
aunger: Line is a little long.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesView.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.wycliffeassociates.otter.jvm.app.ui.resourcetakes.viewmodel.TakesViewModel
+import java.util.EnumMap
+import tornadofx.*
+import org.wycliffeassociates.otter.jvm.utils.getNotNull
+
+class ResourceTakesView : View() {
+    private val viewModel: TakesViewModel by inject()
+    private val tabPane = ChromeableTabPane()
+
+    class ContentTypeToTabMap(map: Map<ContentType, TakesTab?>): EnumMap<ContentType, TakesTab>(map)
```
aunger: Does the value need to be nullable?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesView.kt
```diff
@@ -0,0 +1,70 @@
...
+    private val contentTypeToTabMap = ContentTypeToTabMap(
```
aunger: Subclassing doesn't seem necessary. Can't contentTypeToTabMap just be an EnumMap?

aunger (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesView.kt
KJoslyn: Actually sorry I meant to comment on this. In MapUtils.kt, `getNotNull()` uses reflection to get the class name when throwing an exception. That's really the only reason I subclassed EnumMap.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesView.kt
KJoslyn: Oops, not anymore. It was leftover from before I made the `getNotNull()` function, when nulls were propagating too far.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesStyles.kt
KJoslyn: True, fixing that.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesView.kt
aunger: I might consider just making getNotNull() a method of ContentTypeToTabMap. I'm ok with this as-is, though.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
```diff
@@ -1,43 +1,48 @@
...
-    val chapter: Chapter
-        get() = activeChapterProperty.value
-
-    val activeResourceSlugProperty = SimpleStringProperty()
-    val resourceSlug: String
-        get() = activeResourceSlugProperty.value
+    private val takesViewModel: TakesViewModel by inject()
+    private val workbookViewModel: WorkbookViewModel by inject()
 
     val resourceGroups: ResourceGroupCardItemList = ResourceGroupCardItemList(mutableListOf())
```
jsarabia: If this is a usual use case, perhaps the initialization of an empty mutable list could be done by default and thus not have a parameter here?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesStyles.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    private val topMargin = 15.px
+    private val bottomMargin = 15.px
+    private val leftRegionLeftMargin = 80.px
+    private val leftRegionRightMargin = 100.px
+
+    init {
+        main {
+            prefWidth = Screen.getPrimary().visualBounds.width.px - 600.0
+            prefHeight = Screen.getPrimary().visualBounds.height.px - 200.0
```
jsarabia: Seems like a bad idea, and as a placeholder it will likely be forgotten to be fixed?

KJoslyn: Yeah I don't think this shouldn't be here anymore, I'll remove it

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: Sure, good point

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/TabContentLeftRegion.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.wycliffeassociates.otter.jvm.app.widgets.highlightablebutton.highlightablebutton
+import tornadofx.*
+import tornadofx.FX.Companion.messages
+
+class TabContentLeftRegion(formattedTextProperty: StringProperty, newTakeAction: () -> Unit): VBox() {
+    init {
+        vgrow = Priority.ALWAYS
+        vbox {
+            addClass(ResourceTakesStyles.dragTarget)
+            text("Drag Take Here")
```
jsarabia: messages

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/viewmodel/TakesViewModel.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    val activeRecordableProperty = SimpleObjectProperty<Recordable>()
+    var activeRecordable by activeRecordableProperty
+
+    private val recordTake = RecordTake(
+        WaveFileCreator(),
+        LaunchPlugin(pluginRepository)
+    )
+
+    class ContentTypeToLabelPropertyMap(map: Map<ContentType, SimpleStringProperty?>):
```
jsarabia: Why nullable?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
```diff
@@ -1,43 +1,48 @@
...
     fun loadResourceGroups() {
-        chapter
-            .children
-            .startWith(chapter)
-            .mapNotNull { resourceGroupCardItem(it, resourceSlug, onSelect = this::navigateToTakesPage) }
-            .buffer(2) // Buffering by 2 prevents the list UI from jumping while groups are loading
-            .subscribe {
-                Platform.runLater {
-                    resourceGroups.addAll(it)
+        workbookViewModel.chapter.let { chapter ->
```
jsarabia: why use let here?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
```diff
@@ -1,43 +1,48 @@
...
+                .children
+                .startWith(chapter)
+                .mapNotNull {
+                    resourceGroupCardItem(it, workbookViewModel.resourceSlug, onSelect = this::navigateToTakesPage)
+                }
+                .buffer(2) // Buffering by 2 prevents the list UI from jumping while groups are loading
+                .subscribe {
+                    Platform.runLater {
+                        resourceGroups.addAll(it)
+                    }
```
jsarabia: This should probably happen in the view controller (view/fragment)

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesView.kt
jsarabia: can we avoid nulls altogether here?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: I use `chapter` again at line 23 so I didn't want to write `workbookViewModel.chapter` again, but maybe I should have just done `val chapter = workbookViewModel.chapter` at the start

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/TakesTab.kt
```diff
@@ -0,0 +1,106 @@
...
+    @Suppress("ProtectedInFinal", "Unused")
+    protected fun finalize() {
+        clearDisposables()
+    }
+
+    private fun clearDisposables() {
+        disposables.clear()
+    }
+
+    private fun getFormattedTextBinding() = Bindings.createStringBinding(Callable { getFormattedText() }, recordableProperty)
```
jsarabia: line length

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/viewmodel/TakesViewModel.kt
KJoslyn: Good catch, it shouldn't be nullable

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/view/ResourceTakesView.kt
KJoslyn: Not quite, we can't guarantee that a key exists in the map

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: Isn't the observable list usually contained and manipulated in the view model, and the view/fragment simply listens for changes to the list?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/workbook/viewmodel/WorkbookViewModel.kt
KJoslyn: Update 6/26: Unfortunately I don't think the pattern of throwing an exception on null can apply to every property since Chunk is actually nullable (sometimes we are recording material about a Chapter, not a particular Chunk.)

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/workbook/viewmodel/WorkbookViewModel.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    val activeChapterProperty = SimpleObjectProperty<Chapter?>()
+    val chapter: Chapter
+        get() = activeChapterProperty.value ?: throw IllegalStateException("Chapter is null")
+
+    val activeChunkProperty = SimpleObjectProperty<Chunk?>()
+    val chunk by activeChunkProperty
```
KJoslyn, line 22: See above comment about getters in this file

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resources/viewmodel/ResourcesViewModel.kt
KJoslyn: So it turns out that we don't need to use Platform.runLater when we are adding items to an observable list. The listview will automatically notice changes to the observable list and add the items anyway. I was confused because in the case where resource cards are added to the resource group card, we are actually directly adding a node to the node graph- in which case we need Platform.runLater.

By removing Platform.runLater, we're closer to being able to move this ViewModel into common!

Edit: Unfortunately I'm still a little confused because it seems that I need the Platform.runLater in the loadTakes() function of RecordableTabViewModel. If it isn't there, the listview does not update correctly- I get a duplicate take card showing up when I record a take. I want to understand this!

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/TakeManagementViewModel.kt
KJoslyn: Update 6/26: I removed takesViewModel from here. TakesViewModel actually got renamed to RecordResourceViewModel, which does not contain any recording business logic. TakeManagementViewModel is now the thing that houses logic for recording/editing/etc Takes (and a lot more). 

Likewise, I think we should have a RecordScriptureViewModel that only contains things needed to bind data to the "scripture takes" view. The current TakeManagementViewModel probably needs to be greatly refactored.

jsarabia:
how can we update travis to get the chromable tab pane so the tests work?

jsarabia (approved):

PR #459, opened 2019-06-18
Test to trigger Travis.

aunger:


aunger:
Travis seems to work now.

PR #462, opened 2019-07-01
change 8woc to otter

jsarabia:


KJoslyn (approved):

danparisd:
This PR broke the jenkins auto build. I can fix it, but have other things I'm working on right now. We can leave it broken if you don't need the auto-builds or we could revert this if that has lower impact. Let me know the priority on this so I can get it prioritized on my schedule.

PR #468, opened 2019-07-11
Recordings View Model Refactor

KJoslyn:
Please see the the [RecordableVM](https://www.draw.io/#G1svIhJG1NHo02_sLdSuam9Li89ze4WIX_) page in draw.io to see the new structure of the old "Takes" view models.

This PR does not intend to provide functionality to create, edit, delete, or select for takes.

Requires [common #113](https://github.com/WycliffeAssociates/otter-common/pull/113)

If you want to try drilling down into either the scripture or resources recordings pages, check out branch `kj-temp-navigation-crossmode` and run Main.kt. In the bottom left corner, there is a checkbox that should be selected if you want to simulate being in "Resource Mode". If the checkbox is checked when you select a chapter, the tN resource list page will be loaded for the chapter. If not, verse cards will be displayed instead.

In that test branch, the header is also added at the top so that you can navigate back ONLY between the resource list and resource recordings pages. Otherwise, navigating backwards has undefined behavior.


src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    private fun getChunkList(chunks: Observable<Chunk>) {
+        chunks.toList()
+            .observeOnFx()
+            .subscribe { list ->
+                chunkList.setAll(list.sortedBy { chunk -> chunk.start })
+            }
+    }
+
+    private enum class StepDirection {
```
jsarabia: Move enum declarations to the top

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/view/RecordScriptureFragment.kt
```diff
@@ -58,12 +59,12 @@ class TakeManagementFragment : Fragment() {
             topAnchor = 0.0
         }
         addClass(AppStyles.appBackground)
-        addClass(TakeManagementStyles.tpanelStyle)
+        addClass(RecordScriptureStyles.tpanelStyle)
         val snackBar = JFXSnackbar(this)
-        takeManagementViewModel.snackBarObservable.subscribe { shouldShow ->
+        recordableViewModel.snackBarObservable.subscribe { shouldShow ->
             snackBar.enqueue(
                     JFXSnackbar.SnackbarEvent(messages["noRecorder"], messages["addPlugin"].toUpperCase(), 5000, false, EventHandler {
```
jsarabia: line length

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+import tornadofx.*
+
+class RecordScriptureViewModel : ViewModel() {
+    private val workbookViewModel: WorkbookViewModel by inject()
+    private val audioPluginViewModel: AudioPluginViewModel by inject()
+
+    val recordableViewModel = RecordableViewModel(audioPluginViewModel)
+
+    // This will be bidirectionally bound to workbookViewModel's activeChunkProperty
+    private val activeChunkProperty = SimpleObjectProperty<Chunk?>()
```
jsarabia: does this need to be nullable?
properties can hold nulls, so this has like, two potential levels of nulls?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    fun previousChunk() {
+        stepToChunk(StepDirection.BACKWARD)
+    }
+
+    private fun setHasNextAndPrevious() {
+        if (chunkList.isNotEmpty()) {
+            doSetHasNextAndPrevious()
+        } else {
+            chunkList.isNotEmpty().toProperty().onChangeOnce {
```
aunger: `isNotEmpty` returns a boolean, and I don't think you can just call `toProperty` on that to get updates on the empty status.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+                doSetHasNextAndPrevious()
+            }
+        }
+    }
+
+    private fun doSetHasNextAndPrevious() {
+        if (chunkList.isNotEmpty()) {
+            hasNext.set(activeChunk.start < chunkList.last().start)
+            hasPrevious.set(activeChunk.start > chunkList.first().start)
+        } else throw IllegalStateException("Chunk list is empty")
```
aunger: Would it be ok just to set both to false instead of throwing?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private fun setTitle(chunk: Chunk) {
+        val label = ContentLabel.VERSE.value
```
aunger: i18n?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+        val start = chunk.start
+        title = "$label $start"
+    }
+
+    private fun getChunkList(chunks: Observable<Chunk>) {
+        chunks.toList()
+            .observeOnFx()
+            .subscribe { list ->
+                chunkList.setAll(list.sortedBy { chunk -> chunk.start })
```
aunger: Minor: Move the sorting into a `map` above the observeOn.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
```diff
@@ -0,0 +1,106 @@
...
+                .record(it)
+                .observeOnFx()
+                .doOnSuccess { result ->
+                    showPluginActive = false
+                    when (result) {
+                        RecordTake.Result.NO_RECORDER -> snackBarObservable.onNext(messages["noRecorder"])
+                        else -> {}
+                    }
+                }
+                .subscribe()
```
aunger: combine `doOnSuccess` and `subscribe`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
```diff
@@ -0,0 +1,106 @@
...
+        contextProperty.set(TakeContext.EDIT_TAKES)
+        showPluginActive = true
+        audioPluginViewModel
+            .edit(take)
+            .observeOnFx()
+            .subscribe { result ->
+                showPluginActive = false
+                when (result) {
+                    EditTake.Result.NO_EDITOR -> snackBarObservable.onNext(messages["noEditor"])
+                    else -> {}
```
aunger: Use success instead of else, so any added result possibilities will demand an update here.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
```diff
@@ -0,0 +1,106 @@
...
+            contextProperty.set(TakeContext.RECORD)
+            showPluginActive = true
+            audioPluginViewModel
+                .record(it)
+                .observeOnFx()
+                .doOnSuccess { result ->
+                    showPluginActive = false
+                    when (result) {
+                        RecordTake.Result.NO_RECORDER -> snackBarObservable.onNext(messages["noRecorder"])
+                        else -> {}
```
aunger: Use an exhaustive `when` here instead of `else`; it'll only add one line.

src/main/kotlin/org/wycliffeassociates/otter/jvm/utils/TornadoFXUtils.kt
```diff
@@ -13,3 +17,33 @@ fun <T> ObservableValue<T>.onChangeAndDoNow(op: (T?) -> Unit) {
...
+ */
+fun <T> Observable<T>.observeOnFxSafe(): Observable<T> {
+    return when (IsFXInitialized.isInitialized()) {
+        true -> observeOnFx()
+        false -> this
+    }
+}
+
+private object IsFXInitialized {
+    private var mIsInitialized: Boolean? = null
```
jsarabia: why are you using hungarian notation here?

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
KJoslyn: I thought I replied to this, but maybe not. You're right, we didn't need this to be nullable, so I changed it to non-nullable.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
KJoslyn: You're right, good catch. I changed it to use `chunkList.sizeProperty.onChangeOnce`

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
KJoslyn: It should be fine, yes. Made that change too.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
KJoslyn: That's better, thanks! Done.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/AudioPluginViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val injector: Injector by inject()
+    private val pluginRepository = injector.pluginRepository
+
+    private val workbookViewModel: WorkbookViewModel by inject()
+
+    private val launchPlugin = LaunchPlugin(pluginRepository)
+    private val recordTake = RecordTake(WaveFileCreator(), launchPlugin)
+    private val editTake = EditTake(launchPlugin)
+
+    fun record(recordable: Recordable): Single<RecordTake.Result> = recordTake
```
jsarabia: https://kotlinlang.org/docs/reference/coding-conventions.html#expression-body-formatting

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
KJoslyn: Not sure where this was in the code back then, but it finally hit me that every time I do a force push after I've rebased a PR branch onto an older PR branch that got updated, we lose the old commits so github can't find them. @aunger This hasn't been a pain point so much for me but it likely is for you and @sarabiaj as you're reviewing stacked PR's. Anyway, this comment should be resolved now.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
KJoslyn: done

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
KJoslyn: done

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordScriptureViewModel.kt
KJoslyn: done

jsarabia (changes requested):
There's a formatting comment that applies in a few places in that file.

The comment about hungarian notation is really just meant to say, don't do that. I know you're trying to differentiate between something public and private with the same name, we have *some* precedence in going with _privatename elsewhere in the code, and I'd rather just stick to that.

Otherwise, looks good to me!

src/main/kotlin/org/wycliffeassociates/otter/jvm/utils/TornadoFXUtils.kt
KJoslyn: Besides the fact that I should use an underscore, there are two other ways I could have implemented this. The first is like this (It is not good, I am only including it because it turned into an interesting thought exercise):

```
private object IsFXInitialized {
    private var _isInitialized: Boolean? = null
        get() {
            if (field == null) {
                field = testIsInitialized()
            }
            return field
        }
    fun isInitialized(): Boolean = _isInitialized!!

    private fun testIsInitialized(): Boolean {
        try {
            Platform.runLater{}
        } catch (e: IllegalStateException) {
            return false
        }
        return true
    }
}
```

Obviously we probably don't like this because there's still a naming issue, and now we need a `!!` operator. 

The second way is like this, which is what I ended up going with and committing just now.

```
private object IsFXInitialized {
    // Since object declarations are initialized lazily, isInitialized will be computed when
    // IsFXInitialized is accessed for the first time
    val isInitialized: Boolean = testIsInitialized()

    private fun testIsInitialized(): Boolean {
        try {
            Platform.runLater{}
        } catch (e: IllegalStateException) {
            return false
        }
        return true
    }
}
```

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/AudioPluginViewModel.kt
KJoslyn: Made the changes!

jsarabia (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/resourcetakes/viewmodel/TabRecordableViewModel.kt
```diff
@@ -0,0 +1,23 @@
...
+    audioPluginViewModel: AudioPluginViewModel
+): RecordableViewModel(
+    audioPluginViewModel
+) {
+    fun getFormattedTextBinding(): StringBinding = Bindings.createStringBinding(
+        Callable { getFormattedText() },
+        recordableProperty
+    )
+
+    private fun getFormattedText(): String? = recordable?.textItem?.text
```
jsarabia, line 22: what is recordable? it doesn't look like it's declared?
I see this wasn't a simple rename of TabRecordableViewModel from RecordableTabViewModel, and a lot was removed

KJoslyn: TabRecordableViewModel inherits from RecordableViewModel now, and RecordableViewModel has `recordable`. That's also why a lot was removed- the functionality is in the base class RecordableViewModel

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
aunger: Ah, gotcha.  The "stacked PRs" are never going to be awesome.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
```diff
@@ -0,0 +1,107 @@
...
+            showPluginActive = true
+            audioPluginViewModel
+                .record(it)
+                .observeOnFx()
+                .subscribe { result ->
+                    showPluginActive = false
+                    when (result) {
+                        RecordTake.Result.NO_RECORDER -> snackBarObservable.onNext(messages["noRecorder"])
+                        RecordTake.Result.SUCCESS, RecordTake.Result.NO_AUDIO -> {}
+                        null -> {} // This cannot happen but the compiler complains if null branch does not exist
```
aunger, line 57: As in another review, this is avoidable if the type is specified as non-null for JVM interop. 

Same as there, I don't think this is worth another commit, but fix it if you wish.

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
```diff
@@ -0,0 +1,107 @@
...
+        showPluginActive = true
+        audioPluginViewModel
+            .edit(take)
+            .observeOnFx()
+            .subscribe { result ->
+                showPluginActive = false
+                when (result) {
+                    EditTake.Result.NO_EDITOR -> snackBarObservable.onNext(messages["noEditor"])
+                    EditTake.Result.SUCCESS -> {}
+                    null -> {} // This cannot happen but the compiler complains if null branch does not exist
```
aunger, line 74: As above.

aunger (approved):

src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt
KJoslyn: Perhaps it's not worth another commit, but I was too curious to not try it. I can't get it to work though- do you happen to know how? (I can't apply the `@NotNull` annotation to an enum class)

KJoslyn: Never mind- I figured out that you can specify the type when you name the parameter in the callback. Thanks @aunger!

PR #476, opened 2019-08-18
Integration test suite

aunger:
Create a suite for integration tests.  It can be run in Gradle locally, and Travis is set to run it.  Currently, the only tests import sample RCs and verify the expected number of rows in key database tables.

---
This change is [Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)


.travis.yml
```diff
@@ -19,7 +19,7 @@ install:
 jobs:
   include:
     - stage: build
-      script: 
-        - ./gradlew build
+      script:
+        - ./gradlew build test integrationTest
```
danparisd: Before this PR we were just running `./gradlew build` and the `test` task was still being kicked off. I think this is a result of functionality of the `java` plugin: https://docs.gradle.org/current/userguide/java_plugin.html in that it extends the `build` task to depend on the `check` task, which in turn depends on the `test` task. 

With the explicit call to `test` here Gradle seems to be smart enough to only run the tests once, but this does seem to have some redundancy we should probably resolve, or at least have a better understanding of what's happening here.

We have a section in `build.gradle` for defining task orders (depends), and we seem to be using that for things that we want to run every time we build the app. My opinion is that we continue to use that section to define gradle task orders and depends and that we only call one gradle task from travis. I'm very open to not having this be the case any longer if there is reason to drop this convention.

Thoughts?

jsarabia (changes requested):
> Reviewed 2 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: 2 of 18 files reviewed, 2 unresolved discussions (waiting on @aunger, @danparisd, @KJoslyn, and @sarabiaj)

___
*[build.gradle, line 150 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmdtGAYEaI_ZwwSdE4G:-LmdtGAZDrjaln0uSZuF:bk0fdo6) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/build.gradle#L150)):*
> ```Gradle
>     integrationTestImplementation sourceSets.main.compileClasspath
>     integrationTestImplementation sourceSets.main.output
> //    integrationTestImplementation sourceSets.integrationTest.output
> ```

should this be in or out?
___
*[src/integration-test/kotlin/integrationtest/rcimport/TestDirectoryProviderModule.kt, line 37 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmeD0OE9WcRVdcda2ao:-LmeD0OFBUrHgx-e0EgG:b17dvj9) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/src/integration-test/kotlin/integrationtest/rcimport/TestDirectoryProviderModule.kt#L37)):*
> ```Kotlin
>     override fun getAppDataDirectory(appendedPath: String) = buildDir(app, appendedPath)
> 
>     override fun getProjectAudioDirectory(
> ```

A lot of these implementations seem like they have no external dependencies or platform assumptions; should we just move them into the interface itself?

aunger (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: 0 of 20 files reviewed, 3 unresolved discussions (waiting on @danparisd, @KJoslyn, and @sarabiaj)

___
*[.travis.yml, line 23 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmdsegdvWD2myv2kdng-r1-23:-Lmkmhu45cudnVQcyZGI:b3rf2a8) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/.travis.yml#L23)):*
<details><summary><i>Previously, danparisd wrote…</i></summary><blockquote>

Before this PR we were just running `./gradlew build` and the `test` task was still being kicked off. I think this is a result of functionality of the `java` plugin: https://docs.gradle.org/current/userguide/java_plugin.html in that it extends the `build` task to depend on the `check` task, which in turn depends on the `test` task. 

With the explicit call to `test` here Gradle seems to be smart enough to only run the tests once, but this does seem to have some redundancy we should probably resolve, or at least have a better understanding of what's happening here.

We have a section in `build.gradle` for defining task orders (depends), and we seem to be using that for things that we want to run every time we build the app. My opinion is that we continue to use that section to define gradle task orders and depends and that we only call one gradle task from travis. I'm very open to not having this be the case any longer if there is reason to drop this convention.

Thoughts?
</blockquote></details>

We don't want integration tests to run every build.  I think think having it run every push or PR is good.  I'll combine the build and test/check calls into one here though.
___
*[build.gradle, line 150 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmdtGAYEaI_ZwwSdE4G:-Lmko2mh9N3HrITRS0ac:b4bv94d) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/build.gradle#L150)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

should this be in or out?
</blockquote></details>

Reply hazy.  Taking it out though.
___
*[src/integration-test/kotlin/integrationtest/rcimport/TestDirectoryProviderModule.kt, line 37 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmeD0OE9WcRVdcda2ao:-Lml0s2mCObmZVgMP4uD:br22du3) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/src/integration-test/kotlin/integrationtest/rcimport/TestDirectoryProviderModule.kt#L37)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

A lot of these implementations seem like they have no external dependencies or platform assumptions; should we just move them into the interface itself?
</blockquote></details>

To me, to put implementation into an interface doesn't feel right, even if we don't think the paths will vary from platform to platform.  

I do have an couple ideas that will let the test and main implementation classes share code.

Idea 1: Extract `buildDir` and the path builders (what get passed into `buildDir`) into [strategies](https://en.wikipedia.org/wiki/Strategy_pattern) in a third class (or in the companion object) that both impls of IDirectoryProvider can use in a simple way.

Idea 2: Delete `TestDirectoryProvider` and allow `DirectoryProvider` to optionally accept some path overrides that will be passed in by `TestDirectoryProviderModule`.

Would it be ok to open this as a separate task?  I've already done more refactoring in the production code than I like in a test PR.

danparisd (commented):
> Reviewed 1 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: 1 of 20 files reviewed, 3 unresolved discussions (waiting on @aunger, @danparisd, @KJoslyn, and @sarabiaj)

___
*[.travis.yml, line 23 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmdsegdvWD2myv2kdng-r1-23:-LmzwvR93SAaxvSiAByC:b89nsgn) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/.travis.yml#L23)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

We don't want integration tests to run every build.  I think think having it run every push or PR is good.  I'll combine the build and test/check calls into one here though.
</blockquote></details>

I don't think I follow. We want to run gradle build every push or PR right? Perhaps we need to move integration testing into it's own build stage and put some conditions on that stage? https://docs.travis-ci.com/user/build-stages/ and https://docs.travis-ci.com/user/conditional-builds-stages-jobs/ may be relevant?

aunger (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: 1 of 20 files reviewed, 3 unresolved discussions (waiting on @danparisd, @KJoslyn, and @sarabiaj)

___
*[.travis.yml, line 23 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmdsegdvWD2myv2kdng-r1-23:-Lmv_0ZV93e69xCfQgZW:b-x0xt8v) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/.travis.yml#L23)):*
<details><summary><i>Previously, danparisd wrote…</i></summary><blockquote>

I don't think I follow. We want to run gradle build every push or PR right? Perhaps we need to move integration testing into it's own build stage and put some conditions on that stage? https://docs.travis-ci.com/user/build-stages/ and https://docs.travis-ci.com/user/conditional-builds-stages-jobs/ may be relevant?
</blockquote></details>

We want it to run on every push/pr. We just don't want it to run on every _build_, because we run lots of little builds locally in our development loop.

danparisd (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: 1 of 20 files reviewed, 2 unresolved discussions (waiting on @danparisd, @KJoslyn, and @sarabiaj)

___
*[.travis.yml, line 23 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmdsegdvWD2myv2kdng-r1-23:-LnD2G8_124hRrcJCm1q:bf44ah1) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/.travis.yml#L23)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

We want it to run on every push/pr. We just don't want it to run on every _build_, because we run lots of little builds locally in our development loop.
</blockquote></details>

OK, so we want to run `gradle build` every travis build, but we don't want to make `gradle build` imply integrationTest because you run `gradle build` frequently as part of development. In that case, I think running `gradle build integrationTest` would be what we want travis doing. This will more than likely change soon anyway as I move the full build into travis, so we can discuss more then.

aunger (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: 0 of 20 files reviewed, 2 unresolved discussions (waiting on @danparisd, @KJoslyn, and @sarabiaj)

___
*[.travis.yml, line 23 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmdsegdvWD2myv2kdng-r1-23:-LnDStpKEL0D1Npp0UZ2:bb6un1j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/.travis.yml#L23)):*
<details><summary><i>Previously, danparisd wrote…</i></summary><blockquote>

OK, so we want to run `gradle build` every travis build, but we don't want to make `gradle build` imply integrationTest because you run `gradle build` frequently as part of development. In that case, I think running `gradle build integrationTest` would be what we want travis doing. This will more than likely change soon anyway as I move the full build into travis, so we can discuss more then.
</blockquote></details>

Done

aunger (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: 0 of 20 files reviewed, 2 unresolved discussions (waiting on @danparisd, @KJoslyn, and @sarabiaj)

___
*[src/integration-test/kotlin/integrationtest/rcimport/TestDirectoryProviderModule.kt, line 37 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476#-LmeD0OE9WcRVdcda2ao:-LnDTBTM5EbAoiG_9Zwm:b6vx7i6) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/6fe780e37bb6d1f37573d5e5cb122b856ce75d52/src/integration-test/kotlin/integrationtest/rcimport/TestDirectoryProviderModule.kt#L37)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

To me, to put implementation into an interface doesn't feel right, even if we don't think the paths will vary from platform to platform.  

I do have an couple ideas that will let the test and main implementation classes share code.

Idea 1: Extract `buildDir` and the path builders (what get passed into `buildDir`) into [strategies](https://en.wikipedia.org/wiki/Strategy_pattern) in a third class (or in the companion object) that both impls of IDirectoryProvider can use in a simple way.

Idea 2: Delete `TestDirectoryProvider` and allow `DirectoryProvider` to optionally accept some path overrides that will be passed in by `TestDirectoryProviderModule`.

Would it be ok to open this as a separate task?  I've already done more refactoring in the production code than I like in a test PR.
</blockquote></details>

@sarabiaj So, I went ahead and tried Idea 2 and updated the PR. It's embarrassing how much simpler it made everything.  

Done.

jsarabia (approved):

jsarabia (approved):
> Reviewed 16 of 18 files at r1, 4 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/476)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @KJoslyn)

aunger:
Replaced by #496

PR #499, opened 2019-09-09
Disable back button when back stack is empty

KJoslyn:
The back button is now only enabled when the navigation back stack is not empty. This prevents a crash that would happen when the user would click the back button even though the back stack was emtpy. This PR is in addition to common [#131](https://github.com/WycliffeAssociates/otter-common/pull/131) which actually checks the stack before performing the back() operation.

---
This change is [Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/499)


jsarabia (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/499)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #503, opened 2019-09-10
Jbs in windowed plugins

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)


jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: 0 of 10 files reviewed, 3 unresolved discussions

___
*[build.gradle, line 179 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRr3JJB-0W1DLGktrB:-LoRr3JJB-0W1DLGktrC:by9d4kv) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/build.gradle#L179)):*
> ```Gradle
>         attributes(
>         'Main-Class': "org.wycliffeassociates.otter.jvm.app.MainKt"
>         //'Main-Class': "org.wycliffeassociates.otter.jvm.recorder.app.RecordingAppKt"
> ```

Didn't mean to commit this; will remove
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 53 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrA-C0Zzlm1FCDQaQ:-LoRrA-C0Zzlm1FCDQaR:b-3am651) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L53)):*
> ```Kotlin
>         recordable?.let {
>             contextProperty.set(TakeContext.RECORD)
>             // showPluginActive = true
> ```

Still need to figure out a way around this
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt, line 40 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrIeU0oTw8Iix4ebL:-LoRrIeV0EhKkfRTz_7F:b-6jtj7k) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt#L40)):*
> ```Kotlin
>                 recorderViewModel.onViewReady()
>             }
>         }.start()
> ```

Need to figure out how to tell if the view has been drawn in order to know the width; setOnShown worked when the plugin was launched standalone, but did not appear to work in window, so I temporarily added a Thread sleep here

aunger (changes requested):
> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 10 unresolved discussions (waiting on @aunger and @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 82 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSUAHMCiP5K4Ic3uaY:-LoSUAHMCiP5K4Ic3uaZ:b-gb1o7i) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L82)):*
> ```Kotlin
>         if (insertedArgs.isEmpty()) {
>             insertedArgs.add("--wav=$audioFilePath")
>         }
> ```

Minor: this function would likely be simpler using immutable, functional-style map/filter
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 87 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSUZAxC86dxK1XMKra:-LoSUZAy7ubH60EPM1sL:bksfr2u) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L87)):*
> ```Kotlin
> 
>     private fun buildBinArguments(requestedArgs: List<String>, audioFilePath: String): Array<String> {
>         val insertedArgs = mutableListOf<String>()
> ```

Minor: As above, simpler in functional style
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 107 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSVcZv0hRQivMpc0wJ:-LoSVcZv0hRQivMpc0wK:b-ov88n0) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L107)):*
> ```Kotlin
>         finder.add(jar)
>         val filter = SubclassClassFilter(PluginEntrypoint::class.java)
>         val foundClasses = mutableListOf<ClassInfo>()
> ```

The Clapper library doesn't appear to require mutable collections.  Can you just use the basic immutable overload?
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 113 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSVvpV9aUkV0RaPhws:-LoSVvpV9aUkV0RaPhwt:bkipnph) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L113)):*
> ```Kotlin
>             Reflection.createKotlinClass(pluginClass) as KClass<PluginEntrypoint>
>         } else {
>             null
> ```

untested idea:
```kotlin
val foundClasses = finder.getClasses(filter)
return foundClasses
    .firstOrNull
    ?.let { /* classloader stuff */ }
```
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 117 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSX9VB-C5LmogLPBjW:-LoSX9VB-C5LmogLPBjX:b-c2a909) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L117)):*
> ```Kotlin
>     }
> ```

extra linefeed
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 119 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSXB6A3FEy2GMuapfG:-LoSXB6A3FEy2GMuapfH:b-k2i002) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L119)):*
> ```Kotlin
> 
>     private fun runProcess(audioFile: File, processArgs: List<String>) {
>         // Build and start the process
> ```

Nit: I know it's not your comment, but please remove it since it's pointless
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 129 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSXnyOBgj5q82280UO:-LoSXnyOBgj5q82280UP:b-8y2jqd) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L129)):*
> ```Kotlin
>     }
> 
>     private fun runInWindow(pluginClass: KClass<PluginEntrypoint>, parameters: Parameters) {
> ```

`runInWindow` sounds more like running in a separate window to me. Maybe something like `runInOtterMainWindow` is better?
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt, line 40 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoS_pM5E0IJ0Vtujz5H:-LoS_pM5E0IJ0Vtujz5I:b-64nxxy) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt#L40)):*
> ```Kotlin
>                 recorderViewModel.onViewReady()
>             }
>         }.start()
> ```

You appear to be calling onViewReady twice.  Once in the `if` or `else`, and once in the thread.
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt, line 40 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrIeU0oTw8Iix4ebL:-LoSubmh1lRQE3-nWbsx:b-e75pxc) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt#L40)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

Need to figure out how to tell if the view has been drawn in order to know the width; setOnShown worked when the plugin was launched standalone, but did not appear to work in window, so I temporarily added a Thread sleep here
</blockquote></details>

What do you get if you read the width before it's drawn?  Could you check it a few times a second until you get a reasonable value?

aunger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 10 unresolved discussions (waiting on @aunger and @sarabiaj)

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 10 unresolved discussions (waiting on @aunger and @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 107 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSVcZv0hRQivMpc0wJ:-LoVAYU84ys1xgs7hSNo:bvhl5wh) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L107)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

The Clapper library doesn't appear to require mutable collections.  Can you just use the basic immutable overload?
</blockquote></details>

I don't understand what you mean by immutable overload?
The clapper library is a java library where the standard collections are all mutable. Line 310 of ClassFinder.java contains the following:
``` classes.add (classInfo);```
This is also a private method variable that never leaves a private method.
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt, line 40 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrIeU0oTw8Iix4ebL:-LoVCFiC9dztk5H5Dr0w:b-gd66bo) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt#L40)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

What do you get if you read the width before it's drawn?  Could you check it a few times a second until you get a reasonable value?
</blockquote></details>

If the canvas hasn't been drawn, the width is 0; but I need to make an array in the renderer equal to this width.
Checking a few times a second feels equally hacky; I've always seen a lifecycle callback provided by the UI framework for this kind of thing, so I suspect I just haven't found it yet.

Technically the inability to adjust to changing widths by definition makes this view unresponsive to users resizing the window, which was previously not a concern for this "app" on Android. It feels semi in the realm of new functionality to handle that though, since it means binding a method to properties to pause rendering to reallocate an array and copy all existing values, but also try to make sure that only happens once the dust has settled and the user isn't still resizing the window.
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt, line 40 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoS_pM5E0IJ0Vtujz5H:-LoVDmHP4S3NyPXPSudN:b31ky9l) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt#L40)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

You appear to be calling onViewReady twice.  Once in the `if` or `else`, and once in the thread.
</blockquote></details>

yeah, kind of some debug code I guess. The if/else had been working when running standalone (I assume when the stage is first created). This was what I assumed to be the lifecycle method I needed to attach to, but none of it was called. The thread was as a brute force approach to at least get it working (and also to know that's why previously it wasn't drawing the waveform and I had no idea what was wrong since it was working standalone)

**Review by jsarabia — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: 9 of 10 files reviewed, 7 unresolved discussions (waiting on @aunger)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 82 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSUAHMCiP5K4Ic3uaY:-LoVKslj762MrMDwcNYu:bb6un1j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L82)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

Minor: this function would likely be simpler using immutable, functional-style map/filter
</blockquote></details>

Done
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 87 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSUZAxC86dxK1XMKra:-LoVKuceBZkIptTlSFE7:bb6un1j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L87)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

Minor: As above, simpler in functional style
</blockquote></details>

Done
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 113 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSVvpV9aUkV0RaPhws:-LoVKxCt1TpdUf2TeDbd:bb6un1j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L113)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

untested idea:
```kotlin
val foundClasses = finder.getClasses(filter)
return foundClasses
    .firstOrNull
    ?.let { /* classloader stuff */ }
```
</blockquote></details>

Done
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 117 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSX9VB-C5LmogLPBjW:-LoVKyoT8GM9qWUbspvt:bb6un1j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L117)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

extra linefeed
</blockquote></details>

Done
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 119 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSXB6A3FEy2GMuapfG:-LoVL-Ln5OstNuYQDrrB:bb6un1j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L119)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

Nit: I know it's not your comment, but please remove it since it's pointless
</blockquote></details>

Done
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 129 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSXnyOBgj5q82280UO:-LoVL0Y-5neiFo0ctxME:bb6un1j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L129)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

`runInWindow` sounds more like running in a separate window to me. Maybe something like `runInOtterMainWindow` is better?
</blockquote></details>

Done

aunger (commented):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 5 unresolved discussions (waiting on @aunger and @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 107 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoSVcZv0hRQivMpc0wJ:-LoWMXpDDi4KLLlqZwcG:b3ggd18) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L107)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

I don't understand what you mean by immutable overload?
The clapper library is a java library where the standard collections are all mutable. Line 310 of ClassFinder.java contains the following:
``` classes.add (classInfo);```
This is also a private method variable that never leaves a private method.

</blockquote></details>

Oh, wild. There are two libraries, and I thought you were using the other one:

http://software.clapper.org/javautil/
vs
http://software.clapper.org/classutil/#finding-all-classes-that-implement-an-interface-directly-or-indirectly

This is fine as is.
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt, line 40 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrIeU0oTw8Iix4ebL:-LoWNnN39ZBSIrFKAUbp:b8nls9c) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt#L40)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

If the canvas hasn't been drawn, the width is 0; but I need to make an array in the renderer equal to this width.
Checking a few times a second feels equally hacky; I've always seen a lifecycle callback provided by the UI framework for this kind of thing, so I suspect I just haven't found it yet.

Technically the inability to adjust to changing widths by definition makes this view unresponsive to users resizing the window, which was previously not a concern for this "app" on Android. It feels semi in the realm of new functionality to handle that though, since it means binding a method to properties to pause rendering to reallocate an array and copy all existing values, but also try to make sure that only happens once the dust has settled and the user isn't still resizing the window.
</blockquote></details>

How about passing the available space dimensions in via the scope? Resize won't work, of course; the user will just need to size the window properly before hitting record.  

This approach would spawn a follow-up, low-pri (non-mvp) task to make it responsive to resizing.

**Review by aunger — CHANGES_REQUESTED** [review]
> Reviewed 5 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 5 unresolved discussions (waiting on @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 53 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrA-C0Zzlm1FCDQaQ:-LoX8tGEDYUXCjWX8cwr:b-xbnlxb) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L53)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

Still need to figure out a way around this
</blockquote></details>

Can this be passed to `record` below, and then plumbed through to where you can decide whether to use it?
```kotlin
audioPluginViewModel
    .record(recordable = it, disableOtterUx = { showPluginActive = true })
    .observeOnFx()
    .subscribe { ... }
```
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 69 at r3](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoX8iLS75vnuIqT9SKN:-LoX8iLTC6cjEe2oeGsB:bmjq0t8) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/fde7248af62054841983fe875442beb49c98679f/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L69)):*
> ```Kotlin
>     fun editTake(editTakeEvent: EditTakeEvent) {
>         contextProperty.set(TakeContext.EDIT_TAKES)
>         showPluginActive = true
> ```

I assume this needs to change too
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/viewmodel/RecorderViewModel.kt, line 68 at r3](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoXCYQ39477wmYGCHk9:-LoXCYQ39477wmYGCHkA:b-n2sw0j) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/fde7248af62054841983fe875442beb49c98679f/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/viewmodel/RecorderViewModel.kt#L68)):*
> ```Kotlin
>     }
> 
>     fun onViewReady(width: Int) {
> ```

Should this be idempotent?  It appears to be called every time the RecordingVisualizerFragment is resized.

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 4 unresolved discussions (waiting on @aunger and @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 53 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrA-C0Zzlm1FCDQaQ:-Lo_NmHS3hfa2hXNsjMv:bdw9uh) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L53)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

Can this be passed to `record` below, and then plumbed through to where you can decide whether to use it?
```kotlin
audioPluginViewModel
    .record(recordable = it, disableOtterUx = { showPluginActive = true })
    .observeOnFx()
    .subscribe { ... }
```
</blockquote></details>

yeah, that's probably the best approach
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 69 at r3](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoX8iLS75vnuIqT9SKN:-Lo_Nhy8ENIDvH-ijcJo:b2slxl4) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/fde7248af62054841983fe875442beb49c98679f/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L69)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

I assume this needs to change too
</blockquote></details>

Anywhere a plugin can be called; yes.
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/viewmodel/RecorderViewModel.kt, line 68 at r3](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoXCYQ39477wmYGCHk9:-Lo_MvmjCd5j3qS_805P:b-g6tuhb) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/fde7248af62054841983fe875442beb49c98679f/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/viewmodel/RecorderViewModel.kt#L68)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

Should this be idempotent?  It appears to be called every time the RecordingVisualizerFragment is resized.
</blockquote></details>

..It shouldn't be currently? There's a boolean to make sure it only happens once. It's a stepping stone to getting this working responsively if that's desired though.

aunger (commented):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 5 unresolved discussions (waiting on @aunger and @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 103 at r4](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-Lo_wqbO1xdSBFmdjDTw:-Lo_wqbQCJUR6tQvc-Iv:bp9orgn) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/b4221fdf84f6a65592e61f97cb9f8b9b4fff4939/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L103)):*
> ```Kotlin
>             ?.let { foundClass ->
>                 val pluginClass = javaClass.classLoader.loadClass(foundClass.className)
>                 @Suppress("UNCHECKED_CAST")
> ```

This is fine. You could use `as?` to prevent a possible exception and remove the need for warning suppression, but if the cast fails, you might actually want an exception.

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 5 unresolved discussions (waiting on @aunger)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 53 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrA-C0Zzlm1FCDQaQ:-Loa-E6o0WjhiRnrysOm:b-z8hmvd) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L53)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

yeah, that's probably the best approach
</blockquote></details>

Discussed refactoring the plugin system to move to an enum for recording/editing rather than a long chain of calls and mostly duplicated code
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt, line 103 at r4](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-Lo_wqbO1xdSBFmdjDTw:-Lo_zSgR9Bj0aTomEdb9:bv50zgx) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/b4221fdf84f6a65592e61f97cb9f8b9b4fff4939/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audioplugin/AudioPlugin.kt#L103)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

This is fine. You could use `as?` to prevent a possible exception and remove the need for warning suppression, but if the cast fails, you might actually want an exception.
</blockquote></details>

I'd probably want the exception since if the as? failed... it would be concerning and worth the exception. The firstOrNull not being null means a class was found implementing PluginEntrypoint; so if it wasn't actually implementing that, there would be an issue with the library. If I can't make it into a KClass, it would be an issue with the Reflection library =\
___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt, line 40 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoRrIeU0oTw8Iix4ebL:-Lo_yNji87kwDpKHd4NX:b1t6u76) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/2107a24b59bc9105eea9ac73e190705b56c4d815/src/main/kotlin/org/wycliffeassociates/otter/jvm/recorder/app/view/RecorderView.kt#L40)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

How about passing the available space dimensions in via the scope? Resize won't work, of course; the user will just need to size the window properly before hitting record.  

This approach would spawn a follow-up, low-pri (non-mvp) task to make it responsive to resizing.
</blockquote></details>

The best I could do for the scope is to get the window width at the time that the plugin is called. I could query the window width from within the plugin though.

Really, this width should be smaller than the window width though since it should not include the width of the volume bar. There may be a better solution, but currently I have it bound to the width property which is a stepping stone to making it responsive to resizing. Currently, it sets an initialized flag so as to not call the onViewReady method again, but in the future onViewReady could be renamed and handle reallocating the visualizer queues.

aunger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 3 unresolved discussions (waiting on @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 69 at r3](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoX8iLS75vnuIqT9SKN:-Loa7nYiCkzr_6tFG6tX:b-1hlk5l) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/fde7248af62054841983fe875442beb49c98679f/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L69)):*
<details><summary><i>Previously, sarabiaj (Joe) wrote…</i></summary><blockquote>

Anywhere a plugin can be called; yes.
</blockquote></details>

If you'd rather remove this and include a new/edit flag/enum like you mentioned, that's fine too, of course.

aunger (approved):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503)* status: all files reviewed, 2 unresolved discussions (waiting on @sarabiaj)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt, line 69 at r3](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/503#-LoX8iLS75vnuIqT9SKN:-Loa8qKS41NcAGiq1r7y:b-yjt0pi) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/fde7248af62054841983fe875442beb49c98679f/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/takemanagement/viewmodel/RecordableViewModel.kt#L69)):*
<details><summary><i>Previously, aunger (Russell Aunger) wrote…</i></summary><blockquote>

If you'd rather remove this and include a new/edit flag/enum like you mentioned, that's fine too, of course.
</blockquote></details>

LGTM 
Oh, I just heard that edit is deferred because the plugin doesn't do that yet.  This is resolved then.

PR #508, opened 2019-09-16
Move build to travis and update to java11 jre

danparisd:
Adds install4j building to travis and updates the installer, bundled jre's and build to java 11.

---
This change is [Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/508)


build.gradle
```diff
@@ -64,6 +65,10 @@ configurations.all {
     resolutionStrategy.cacheChangingModulesFor 0, 'seconds'
 }
 
+configurations.runtimeClasspath.resolutionStrategy.dependencySubstitution {
+    substitute module('org.jooq:jooq') with module("org.jooq:jooq:$jooq_version")
+}
```
jsarabia, line 70: what is this for?

danparisd: Russell and I were getting this jooq error when running the gui fat jar, and he was able to fix it with that.

`Caused by: java.lang.VerifyError: class jooq.tables.AudioPluginEntity overrides final method org.jooq.impl.AbstractTable.fieldsRow()Lorg/jooq/Row;`

Per Russell:
> I actually fixed the jooq one. There was another error hiding right behind it though. My fix for the first one was to force a version on the jooq dependency.

danparisd: For clarity's sake, the error hiding behind it was the platform dependent javafx stuff that is also fixed in this PR.

build.gradle
```diff
@@ -99,9 +104,6 @@ allprojects {
 
 configurations {
     sqllite
-}
```
danparisd, line 102: This gave me a red squiggle in the gradle plugin for VS code since it was duplicated. So I removed the dupe. If there was a reason for the dupe then I can revert this.

.travis.yml
```diff
@@ -1,25 +1,64 @@
...
+- wget --no-verbose "https://otter-build-nightly.s3.amazonaws.com/linux-amd64-11.0.4.tar.gz"
 install:
   - git clone https://github.com/WycliffeAssociates/chromeabletabpane.git ../chromeabletabpane
   - git clone --branch=dev https://github.com/WycliffeAssociates/kotlin-resource-container.git ../kotlin-resource-container
   - git clone https://github.com/WycliffeAssociates/otter-common.git ../otter-common
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
   - echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, PR=$PR, BRANCH=$BRANCH"
   - source init_common.sh
+# Grab the version from the version file for use later and concatenate the Travis build number to it
+  - if [ -f version.txt ]; then export VERSIONVAR="$(cat version.txt).$TRAVIS_BUILD_NUMBER"; fi
```
aunger: Renamed

build.gradle
```diff
@@ -185,17 +192,26 @@ jar {
...
-        installDir = file('/opt/install4j7')
-        //license must be passed via gradlew commandline -P as it can't be commit to public source control
-        if(project.hasProperty('gradlewCommandLicenseProp'))
-        {
-          license = gradlewCommandLicenseProp
-        }
+    //pass in the install4j directory via -P if it's not in the default install location, e.q. for use with travis or arbitrary paths
+    if(project.hasProperty('gradlewInstall4jDirectory'))
+    {
+    installDir = file(gradlewInstall4jDirectory)
```
aunger: Nit: indentation

build.gradle
```diff
@@ -185,17 +192,26 @@ jar {
...
+    }
+    //license must be passed via gradlew commandline -P as it can't be commit to public source control
+    if(project.hasProperty('gradlewCommandLicenseProp'))
+    {
+    license = gradlewCommandLicenseProp
```
aunger: Nit: indentation

build.gradle
```diff
@@ -185,17 +192,26 @@ jar {
...
+    }
 }
 
 
 task install4jdeploy(type: com.install4j.gradle.Install4jTask) {
     projectFile = 'otter.install4j'
+    //Pass Travis Build ID to install4j for versioning
+    if(project.hasProperty('gradlewCommandVersionProp'))
+    {
+    release = gradlewCommandVersionProp
```
aunger: Nit: indentation

aunger (changes requested):

.travis.yml
```diff
@@ -1,25 +1,64 @@
...
+- wget --no-verbose "https://otter-build-nightly.s3.amazonaws.com/linux-amd64-11.0.4.tar.gz"
 install:
   - git clone https://github.com/WycliffeAssociates/chromeabletabpane.git ../chromeabletabpane
   - git clone --branch=dev https://github.com/WycliffeAssociates/kotlin-resource-container.git ../kotlin-resource-container
   - git clone https://github.com/WycliffeAssociates/otter-common.git ../otter-common
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
   - echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, PR=$PR, BRANCH=$BRANCH"
   - source init_common.sh
+# Grab the version from the version file for use later and concatenate the Travis build number to it
+  - if [ -f VERSION ]; then export VERSIONVAR="$(cat VERSION).$TRAVIS_BUILD_NUMBER"; fi
```
aunger, line 33: What happens if someone unknowingly commits VERSION with e.g. a trailing linefeed? Does the build actually fail, or does it "succeed" (according to the Travis status) with mangled filenames?

Sorry for not mentioning this earlier.

I'm not blocking the pr on this, but here are some ideas if you decide this is a problem: https://stackoverflow.com/questions/369758/how-to-trim-whitespace-from-a-bash-variable

build.gradle
danparisd: Open to suggestions. I'm not good with that sort of stuff.

.travis.yml
danparisd: a line feed does not affect it based on testing. Letters work. Symbols don't all work perfectly but they don't break the build. If install4j fails to build because it doesn't like what is passed in, then it will fail the build. If the upload to S3 doesn't like the filename the upload will fail, but this will not fail the build.

build.gradle
aunger: Pretty sure I added this dupe.  Pretty sure there's no good reason.

build.gradle
aunger: Make the lines between a pair of brackets indented four spaces more than the lines outside of them.  Change this:
```bash
    if(project.hasProperty('gradlewInstall4jDirectory'))
    {
    installDir = file(gradlewInstall4jDirectory)
    }
```
... to this:
```bash
    if(project.hasProperty('gradlewInstall4jDirectory'))
    {
        installDir = file(gradlewInstall4jDirectory)
    }
```

.travis.yml
aunger: Thanks for testing. That's better than I feared. I'm good with this as-is.

danparisd: Cool. Thanks for bringing it to my attention.

build.gradle
danparisd: done. Thanks.

aunger (approved):

PR #512, opened 2019-09-19
Ns small wizard fixes

natibekele:
small pr to fix dirty state in wizard and fix filtering on source languages

---
This change is [Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/512)

### DIFF
--- a/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectwizard/view/fragments/SelectLanguage.kt
+++ b/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectwizard/view/fragments/SelectLanguage.kt
@@ -14,7 +14,7 @@ class SelectLanguage : Fragment() {
         }
         searchField.promptText = messages["languageSearchPrompt"]
         autoSelect = true
-        filter(viewModel::filterLanguages)
+        filter(viewModel::filterSourceLanguages)
         viewModel.clearLanguages.subscribe {
             searchField.clear()
             listView.selectionModel.clearSelection()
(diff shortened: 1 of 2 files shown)


jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/512#-:-LpABSM878LupgKHCYBk:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @sarabiaj)

aunger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/512)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @natibekele)

___
*[src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectwizard/viewmodel/ProjectWizardViewModel.kt, line 204 at r1](https://reviewable.io/reviews/wycliffeassociates/otter-jvm/512#-LpdyqJN2XDYxVoYEZoi:-LpdyqJN2XDYxVoYEZoj:bfljxti) ([raw file](https://github.com/wycliffeassociates/otter-jvm/blob/cb8c4a71910fcf0e8e13f7d355f6409a61dfbad9/src/main/kotlin/org/wycliffeassociates/otter/jvm/app/ui/projectwizard/viewmodel/ProjectWizardViewModel.kt#L204)):*
> ```Kotlin
> 
>     fun filterSourceLanguages(query: String): ObservableList<Language> =
>         filterLanguages(query).filtered { filteredLanguages.contains(it) }
> ```

You already have a function with this name and a different signature. Two overloaded functions should in some sense "do the same thing", but here they don't.

The names are further confusing because with this PR you're starting to differentiate between an ObservableList created here by `filterSourceLanguages` and some other OL created by `filterLanguages`.  However, as I trace the assignments backward, I find that that other filtered list (`filteredLanguages`, the one you're _not_ calling "source") is created by the overloaded `filterSourceLanguages` function. Please rework the variable and function names so "source" always means the same thing.

